PR #47, opened 2024-01-19
Anvil Test Wrapper

nitaliano:
**Description**

* Adds a `test-utils` directory under `op-app`. This includes wrappers around `anvil` and `wagmi` to use in `vitest`
* Adds hook unit tests for everything in `op-app`


packages/op-app/src/test-utils/react.ts
```diff
@@ -0,0 +1,88 @@
...
+  ],
+  pollingInterval: 100,
+  storage: null,
+  transports: {
+    [l1.id]: http(),
+    [l2.id]: http(),
+  },
+})
+
+// needed to preserve WagmiConfig prototype
```
nitaliano, line 33: when you deconstruct an object its functions on the prototype are not copied over

jakim929 (approved):
this is great! 

one high level q:
it seems like we run a single anvil instance (1 for L1, 1 for L2), how would we handle isolating state between tests? do we create separate proxies per test?

nitaliano:
@jakim929 the tests all run in serial so multiple tests will never be hitting the same proxy at the same time

The test client that we create is able to set and reset state, the `afterAll` hook is handling resetting the client per file, if we wanted it can be updated to reset for each test `afterEach`. In order to use the state that you want there is a function that exists on the test client to load it https://viem.sh/docs/actions/test/loadState, so we'll probably have a light wrapper around this in the future.

PR #52, opened 2024-01-25
build(monorepo): add Dockerfile for multi-stage builds

jakim929:
Adds a Dockerfile that supports multi-stage builds. Currently only builds the `paymaster-proxy` app, but the setup will be useful when we begin building/uploading other app images in this monorepo.

- updated `nx` version
- adds `.dockerignore` file
- uses minimal `node:18-alpine` image and installs additional packages (`python3`, `make`, `g++`) to be able to use `node-gyp`
- uses `[pnpm fetch](https://pnpm.io/cli/fetch)` to fetch packages based on lockfile to minimize cache misses
- uses `[pnpm deploy](https://pnpm.io/cli/deploy)` to create portable app that includes all node_modules (including dependencies from the monorepo)


jakim929:
* `main`
  * **PR #52** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/52?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/52?utm_source=stack-comment).

apps/paymaster-proxy/package.json
```diff
@@ -21,5 +21,8 @@
   },
   "dependencies": {
     "express": "^4.18.2"
-  }
+  },
+  "files": [
+    "dist"
+  ]
```
jakim929, line 27: This is required for `pnpm deploy` to include the built `dist` folder, as it currently doesn't include files included in a packages's local `.gitignore`

context: https://github.com/pnpm/pnpm/issues/7286

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# required dependencies for node-gyp
+# context: https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#node-gyp-alpine
+RUN apk add --no-cache python3 make g++
+
+# https://pnpm.io/cli/fetch
+# Fetches packages based on lockfile not package.json - cache is valid even if package.json changes
+RUN pnpm fetch
+
+# install monorepo dependencies, from the virtual store
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --offline
```
jakim929, line 19: `pnpm fetch` isn't helpful right now since we are only building one app. But this may become useful as we add more apps that may require copying over files in the build process that break the cache.

context: https://pnpm.io/cli/fetch

Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM node:18-alpine AS base
```
nitaliano, line 1: this is probably something we want to keep track of with actions but it would be cool if we only rebuild certain layers and publish them to dockerhub only if there was a change in it's code tree

.dockerignore
```diff
@@ -0,0 +1,6 @@
+node_modules
+.git
+.gitignore
+*.md
+dist
+dev-dist
```
nitaliano, line 6: some other ideas for things we can remove

* `*.spec.ts` files i don't think we need to copy any of the tests over
* maybe just a good idea to add permutations of `.env` files just to prevent anything crazy from happening

nitaliano (approved):

.dockerignore
jakim929: good point! will add

Dockerfile
jakim929: that would be great,

for starters, thinking that inside actions we can
- use https://nx.dev/nx-api/nx/documents/affected to decide which targets to build and push

to only rebuild certain layers
- right now it builds everything in the builder step - need to figure out if there's a way to only build packages relevant to the target

PR #68, opened 2024-01-31
build(paymaster-proxy): pull in certs in Dockerfile

jakim929:
- mounts extra-ca-certificates.cert to `usr/local/share/ca-certificates/extra-ca-certificates.crt`
- passes the extra certificates path to env so node picks it up
- fine if the file doesn't exist, app still runs as errors are ignored (warning is emitted though like the following)

```
Warning: Ignoring extra certs from `/usr/local/share/ca-certificates/extra-ca-certificates.crt`, load failed: error:80000002:system library::No such file or directory
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,13 +42,11 @@ RUN pnpm deploy --filter=paymaster-proxy --prod /prod/paymaster-proxy
 ########################################
 
 FROM base AS paymaster-proxy
+
 COPY --from=builder /prod/paymaster-proxy /prod/paymaster-proxy
 WORKDIR /prod/paymaster-proxy
 
-# setup certs for internal services
-RUN update-ca-certificates
-
-ENV NODE_EXTRA_CA_CERTS=/etc/ssl/certs/ca-certificates.crt
+ENV NODE_EXTRA_CA_CERTS=/usr/local/share/ca-certificates/extra-ca-certificates.crt
 
 EXPOSE 7310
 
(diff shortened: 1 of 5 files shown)


jakim929:
* `main`
  * **PR #68** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/68?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/68?utm_source=stack-comment).

Dockerfile
```diff
@@ -44,5 +44,12 @@ RUN pnpm deploy --filter=paymaster-proxy --prod /prod/paymaster-proxy
 FROM base AS paymaster-proxy
 COPY --from=builder /prod/paymaster-proxy /prod/paymaster-proxy
 WORKDIR /prod/paymaster-proxy
+
+# setup certs for internal services
+RUN update-ca-certificates
```
tremarkley, line 49: does this call need to go after `ENV NODE_EXTRA_CA_CERTS=/etc/ssl/certs/ca-certificates.crt`? I know that is the order we do it in gateway docker: https://github.com/ethereum-optimism/gateway/commit/af609004c08b99b09fedea359786d135b523e58c

tremarkley (approved):

Dockerfile
nitaliano: * `update-ca-certificates` will look under `/usr/local/share/ca-certificates` and concatenate the certs it finds into `/etc/ssl/certs/ca-certificates.crt`
* `NODE_EXTRA_CA_CERTS` tells node where to find any of our custom certs, since node ships with it's own set of certs it doesn't use `/etc/ssl/certs/ca-certificates.crt` by default. This value needs to be set before we execute the `pnpm start` command since that is when node will be looking for the env var

nitaliano (approved):

PR #71, opened 2024-02-02
feat(paymaster-proxy): use redis cache for rate limiting

jakim929:
- adds rate limiting using a redis cache that is shared across pods
    - redis client is initialized before server starts listening
    - adds `REDIS_URL` connection string envvar
- adds `rate-limit-redis` and `ioredis` packages
- refactored docker-compose files from docker-compose.base.yml
- created a docker-compose.replicas.yml that brings up multiple pods + traefik reverse proxy to test that the rate limiting works across multiple pods

tested that rate limiting logic works even when requests get load balanced to different pods


apps/paymaster-proxy/docker-compose.replicas.yml
```diff
@@ -0,0 +1,29 @@
...
+      service: redis-app
+    ports:
+      - 6379:6379
+
+  paymaster-proxy-app:
+    extends:
+      file: docker-compose.base.yml
+      service: paymaster-proxy-app
+    labels:
+      - 'traefik.http.routers.paymaster-proxy.rule=Host(`0.0.0.0`)'
```
jakim929, line 16: using this host to make it not require dns configs to test with artillery

apps/paymaster-proxy/docker-compose.base.yml
```diff
@@ -0,0 +1,27 @@
+version: '3.8'
```
nitaliano, line 1: i think it would be better to stick to a single `docker-compose` file and if you need duplication you can use variables like what i'm doing in the gateway repo, extends is really a relic before yaml variables were a thing

apps/paymaster-proxy/docker-compose.base.yml
```diff
@@ -0,0 +1,27 @@
...
+      timeout: 10s
+      retries: 3
+      start_period: 2s
+
+  paymaster-proxy-app:
+    build:
+      context: ../../
+      dockerfile: Dockerfile
+      target: paymaster-proxy
+    environment:
```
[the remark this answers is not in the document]
jakim929: sounds good - will update to this once we start using internal services

apps/paymaster-proxy/package.json
```diff
@@ -10,7 +10,11 @@
     "lint": "eslint \"**/*.{ts,tsx}\" && pnpm prettier --check \"**/*.{ts,tsx}\"",
     "lint:fix": "eslint \"**/*.{ts,tsx}\" --fix --quiet && pnpm prettier \"**/*.{ts,tsx}\" --write --loglevel=warn",
     "generate-example-env": "ts-node scripts/generateExampleEnv.ts",
-    "run:loadtest": "pnpm dlx artillery run loadtest/basic-loadtest.yml"
+    "run:loadtest": "pnpm dlx artillery run loadtest/basic-loadtest.yml",
+    "docker:services:up": "docker compose up -d redis",
+    "docker:up": "docker compose up -d redis paymaster-proxy-app",
+    "docker:replicas:up": "docker compose up -d redis paymaster-proxy-app-replica reverse-proxy",
+    "docker:down": "docker-compose down"
```
nitaliano, line 17: this should probably be `docker compose down`, we should avoid using `docker-compose` since this is the old way

apps/paymaster-proxy/src/middlewares/getRateLimiter.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const getRateLimiter = (redisClient: Redis) =>
+  rateLimit({
+    windowMs: 5 * 60 * 1000, // 5 minutes
+    limit: 100,
+    standardHeaders: 'draft-7',
+    legacyHeaders: false,
+    store: new RedisStore({
+      // see https://github.com/express-rate-limit/rate-limit-redis
+      // @ts-expect-error - Known issue: the `call` function is not present in @types/ioredis
```
nitaliano, line 13: i know this is from the issue theres probably ways around this in the future, i think the args basically get lumped into this  for ioredis `[command, key, value]` where value can be anything

nitaliano (approved):

apps/paymaster-proxy/package.json
jakim929: good point - sry forgot to push this change, will get merged in with a future PR

PR #107, opened 2024-02-08
Implement support section UI

imranjami:
### Description
Implementing the support section UI

Brought in `cva` for introducing a new variant to the `Tile` component. This is how shadcn components are written so I'm laying out this pattern to stay consistent. 

Screenshot 2024-02-08 at 9 31 07 AM


imranjami:
* `main`
  * **PR #107** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/107?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/107?utm_source=stack-comment).

apps/dapp-console/app/components/Tile/Tile.tsx
```diff
@@ -5,32 +5,73 @@ import {
   CardHeader,
 } from '@eth-optimism/ui-components/src/components/ui/card'
 import { Text } from '@eth-optimism/ui-components/src/components/ui/text'
+import { cn } from '@eth-optimism/ui-components/src/lib/utils'
```
tremarkley, line 8: nit: do you think we should give this a different name? hard to tell what cn means

tremarkley (approved):

apps/dapp-console/app/components/Tile/Tile.tsx
imranjami: this is directly from shadcn so i'd rather not change the name to keep consistency. `cn` here is referring to classname just abbreviated.

PR #126, opened 2024-02-15
feat(paymaster-proxy): add validation and proxying logic

jakim929:
## Overview

This PR adds validation and logic to send transactions to an underlying paymaster service, and correctly handling error states (relaying some of them to the user)

A chain's paymaster route will be at `/${chain.id}` (ie, Sepolia is at `/11155111`)

half or the lines changed are from the pnpm-lock change

## High level flow

1. apply rate limits
2. validate the JSON RPC request - just whether the request object fulfills JSON RPC spec
    - if invalid, return JSON RPC error (this is still HTTP 200)
3. validate whether the request params are valid - 1) method is a supported one, and 2) that the correct parameters for the method was provided
    - if invalid, return JSON RPC error (this is still HTTP 200)
4. send transaction to underlying paymaster provider (Alchemy in the initial implementation, but `getPaymasterConfig` is meant to support other providers on unsupported chains)
5. return paymaster response to user
    - if error, check if the error should be a user facing one or not to decide whether to relay


jakim929:
* `main`
  * **PR #126** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/126?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/126?utm_source=stack-comment).

apps/paymaster-proxy/src/errors/PaymasterError.ts
```diff
@@ -0,0 +1,22 @@
+// These are the errors that can be proxied and returned to the user
```
jakim929, line 1: These objects will capture the errors that an underlying paymaster service can return. Right now it's pretty simple, but in the future we can add more detailed cases.

apps/paymaster-proxy/src/errors/JsonRpcError.ts
```diff
@@ -0,0 +1,78 @@
+import type { PaymasterRpcError } from '@/errors/PaymasterError'
```
jakim929, line 1: This is the error that our proxy service will return. In most cases, it just returns the same error that the underlying service returned.

apps/paymaster-proxy/src/helpers/processSingleOrMultiple.ts
```diff
@@ -0,0 +1,11 @@
+// if input is array, returns result[], otherwise returns result
```
jakim929, line 1: JSON RPC supports batched and single requests like

```
batched
--> [
        {"jsonrpc": "2.0", "method": "sum", "params": [1,2,4], "id": "1"},
        {"jsonrpc": "2.0", "method": "notify_hello", "params": [7]},
        {"jsonrpc": "2.0", "method": "subtract", "params": [42,23], "id": "2"},
        {"foo": "boo"},
        {"jsonrpc": "2.0", "method": "foo.get", "params": {"name": "myself"}, "id": "5"},
        {"jsonrpc": "2.0", "method": "get_data", "id": "9"} 
    ]
<-- [
        {"jsonrpc": "2.0", "result": 7, "id": "1"},
        {"jsonrpc": "2.0", "result": 19, "id": "2"},
        {"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}, "id": null},
        {"jsonrpc": "2.0", "error": {"code": -32601, "message": "Method not found"}, "id": "5"},
        {"jsonrpc": "2.0", "result": ["hello", 5], "id": "9"}
    ]


single
--> {"jsonrpc": "2.0", "method": "subtract", "params": {"subtrahend": 23, "minuend": 42}, "id": 3}
<-- {"jsonrpc": "2.0", "result": 19, "id": 3}

```

apps/paymaster-proxy/src/paymaster/types.ts
```diff
@@ -0,0 +1,35 @@
...
+    paymasterAndData: Hex
+    callGasLimit: Hex
+    verificationGasLimit: Hex
+    preVerificationGas: Hex
+  }>
+>
+
+export type PaymasterConfig<T extends Chain = Chain> = {
+  chain: T
+  sponsorUserOperation: SponsorUserOperationImpl
```
jakim929, line 34: in the future we add other operation we support here

apps/paymaster-proxy/src/initializeApiServer.ts
```diff
@@ -20,5 +24,20 @@ export const initializeApiServer = async (
     res.json({ ok: true })
   })
 
+  app.use(express.json(), ((err, req, res, next) => {
```
jakim929, line 27: middleware to catch req.body parsing errors and return

apps/paymaster-proxy/src/helpers/processSingleOrMultiple.ts
```diff
@@ -0,0 +1,11 @@
...
+export const processSingleOrMultiple = async <T, K>(
+  maybeArr: T | T[],
+  fn: (params: T) => Promise<K>,
+): Promise<K | K[]> => {
+  if (Array.isArray(maybeArr)) {
```
nitaliano, line 6: this function can become a single line

```
return Array.isArray(maybeArr) ? Promise.all(maybeArr.map(fn)) : fn(maybeArr)
```

apps/paymaster-proxy/src/initializeApiServer.ts
nitaliano: would turn this into a function with a name so it's more clear what this is doing, would even recommend creating a middleware directory and just throw them all in there

apps/paymaster-proxy/src/ProxyService.ts
```diff
@@ -16,7 +19,20 @@ export class ProxyService {
...
+        policyId: envVars.ALCHEMY_GAS_MANAGER_POLICY_ID_OP_SEPOLIA,
+      }),
+      getAlchemyPaymasterConfig({
+        chain: optimismSepolia,
+        rpcUrl: envVars.ALCHEMY_RPC_URL_OP_SEPOLIA,
+        policyId: envVars.ALCHEMY_GAS_MANAGER_POLICY_ID_OP_SEPOLIA,
+      }),
+    ] as const satisfies PaymasterConfig[]
+
+    const apiServer = await initializeApiServer(redisClient, paymasterConfigs)
```
nitaliano, line 35: it might be better to pass an object into this, i imagine we're going to add much more to this overtime like loggers, sentry, etc

apps/paymaster-proxy/src/rpc/getJsonRpcRequestHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+import { wrapPaymasterResponseIntoJsonRpcResponse } from '@/rpc/wrapPaymasterResponseIntoJsonRpcResponse'
+import type { SupportedPaymasterRequest } from '@/schemas/supportedPaymasterRequestSchema'
+
+const handlePaymasterMethod = async <
+  T extends SupportedPaymasterRequest = SupportedPaymasterRequest,
+>(
+  request: T,
+  sponsorUserOperation: SponsorUserOperationImpl,
+) => {
+  switch (request.method) {
```
nitaliano, line 15: it might be clear for this to become if/else statements if any more logic gets added here, generally think its a good idea to avoid a switch statement if you don't need the default

i also wonder if it's possible to have a map of some sort that basically maps method name to handler

apps/paymaster-proxy/src/rpc/wrapPaymasterResponseIntoJsonRpcResponse.ts
```diff
@@ -0,0 +1,24 @@
+import { JsonRpcError } from '@/errors/JsonRpcError'
+import { PaymasterRpcError } from '@/errors/PaymasterError'
+import type { PaymasterResponse } from '@/paymaster/types'
+
+export const wrapPaymasterResponseIntoJsonRpcResponse = <T>(
+  response: PaymasterResponse<T>,
+  id: number,
+) => {
+  if (response.success === false) {
```
nitaliano, line 9: we can probably turn all instances of `response.success === false` to `!response.success`

apps/paymaster-proxy/src/initializeApiServer.ts
```diff
@@ -20,5 +24,24 @@ export const initializeApiServer = async (
...
+      return res.json(JsonRpcError.parseError().response())
+    }
+
+    next(err)
+  }) as ErrorRequestHandler)
+
+  for (const { chain, sponsorUserOperation } of paymasterConfigs) {
+    console.info(`Registering paymaster route at /${chain.id}: ${chain.name}`)
+
+    app.post(`/${chain.id}`, getJsonRpcRequestHandler({ sponsorUserOperation }))
```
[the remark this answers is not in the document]
jakim929: good point - will namespace under v1 and also /rpc

apps/paymaster-proxy/src/rpc/getJsonRpcRequestHandler.ts
jakim929: will do this in a subsequent PR once we have more than one method

apps/paymaster-proxy/src/rpc/getJsonRpcRequestHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+          )
+
+          return wrapPaymasterResponseIntoJsonRpcResponse(
+            handlePaymasterRequestResult,
+            validationResult.data.id,
+          )
+        },
+      )
+      return res.json(result)
+    } catch (e) {
```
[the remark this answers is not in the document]
jakim929: yep! will explicitly add loggers before we launch

nitaliano (approved):

PR #129, opened 2024-02-20
feat(paymaster-proxy): screen sanctioned addresses

jakim929:
### Overview
- adds support for screening sanctioned addresses using internal screening service
- adds msw to help with mocking fetch responses for testing
- updates docker compose to use the local .env file

### Testing
- tested inside local docker container to check that certs work correctly - using a sanctioned address in https://www.treasury.gov/ofac/downloads/sdn_comments.csv

### DIFF
--- a/apps/paymaster-proxy/src/helpers/screenAddress.ts
+++ b/apps/paymaster-proxy/src/helpers/screenAddress.ts
@@ -18,7 +18,7 @@ const screeningServiceResponse = createJsonRpcResponseSchema(
 )
 
 export const screenAddress = async (address: Address) => {
-  const res = await fetch(`${envVars.SCREENING_SERVICE_URL}`, {
+  const res = await fetch(envVars.SCREENING_SERVICE_URL, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
@@ -34,7 +34,11 @@ export const screenAddress = async (address: Address) => {
   const json = await res.json()
   const response = screeningServiceResponse.parse(json)
 
-  if ('error' in response || response.result.length === 0) {
+  if (
+    'error' in response ||
+    response.result.length === 0 ||
+    !!response.result[0].Error
+  ) {
     throw new Error('Failed to call screening service')
   }
 
--- a/apps/paymaster-proxy/src/helpers/screenAddress.ts
+++ b/apps/paymaster-proxy/src/helpers/screenAddress.ts
@@ -1,0 +1,42 @@
+import type { Address } from 'viem'
+import { z } from 'zod'
+
+import { envVars } from '@/envVars'
+import { addressSchema } from '@/schemas/addressSchema'
+import { createJsonRpcResponseSchema } from '@/schemas/createJsonRpcResponseSchema'
+
+const screeningServiceResultSchema = z.array(
+  z.object({
+    Addr: addressSchema,
+    IsSanctioned: z.boolean(),
+    Error: z.string(),
+  }),
+)
+
+const screeningServiceResponse = createJsonRpcResponseSchema(
+  screeningServiceResultSchema,
+)
+
+export const screenAddress = async (address: Address) => {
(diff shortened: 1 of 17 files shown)


jakim929:
* **#129** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/129?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/129?utm_source=stack-comment).

apps/paymaster-proxy/src/testUtils/mswServer.ts
```diff
@@ -0,0 +1,22 @@
+import { http, HttpResponse } from 'msw'
+import { setupServer } from 'msw/node'
+
+export const mswServer = setupServer()
+
+export const mockPostSuccessJson = (route: string, response: any) => {
```
jakim929, line 6: these handlers should be short-lived, cleared after each test

apps/paymaster-proxy/src/helpers/screenAddress.ts
```diff
@@ -0,0 +1,42 @@
...
+    Error: z.string(),
+  }),
+)
+
+const screeningServiceResponse = createJsonRpcResponseSchema(
+  screeningServiceResultSchema,
+)
+
+export const screenAddress = async (address: Address) => {
+  const res = await fetch(`${envVars.SCREENING_SERVICE_URL}`, {
```
nitaliano, line 21: this can just be `envVars.SCREENING_SERVICE_URL` it doesn't need to wrapped

apps/paymaster-proxy/src/helpers/screenAddress.ts
```diff
@@ -0,0 +1,42 @@
...
+      method: 'svc_screening',
+      params: [[address]],
+      id: 1,
+    }),
+  })
+
+  const json = await res.json()
+  const response = screeningServiceResponse.parse(json)
+
+  if ('error' in response || response.result.length === 0) {
```
nitaliano, line 37: as a follow on it's going to be worth keeping track of when the screening service is failing so it's easier to understand whats happening when we have dashboards setup

some things to think about are making sure we are logging errors, and adding metrics to increment

nitaliano (approved):

apps/paymaster-proxy/src/helpers/screenAddress.ts
tremarkley, line 1: would it make sense to move this helper out of the `paymaster-proxy` app and into a directory that can be shared across multiple apps?

apps/paymaster-proxy/src/helpers/screenAddress.ts
tremarkley, line 37: does this also need to check the response for an error `!!response.result[0].Error`?

tremarkley (approved):

apps/paymaster-proxy/src/helpers/screenAddress.ts
jakim929: yeah good point - as a follow on i will set up logging

apps/paymaster-proxy/src/helpers/screenAddress.ts
jakim929: definitely should, but they don't have a convenient place to live right now

maybe we can discuss how to structure shared packages? also want to avoid pre-refactoring (rule of three) until it's actually used

apps/paymaster-proxy/src/helpers/screenAddress.ts
jakim929: great point! will check for that as well

apps/paymaster-proxy/src/helpers/screenAddress.ts
tremarkley: sounds good, let's talk about how to structure shared packages

PR #132, opened 2024-02-21
event-log-indexer docker & docker-compose setup

nitaliano:
**Description**

* Updates `Dockerfile` to generate ponders image
* Adds docker compose setup to spin up a local postgres, and new image locally
* Adds `setup-local-db.sql` to mimic having `read_only` and `read_write` roles in our deployed GCP dbs and hopefully avoiding permission errors when ponder attempts to create a new schema
* Checks in generated graphql schema so we know exactly what schema each image version is running

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN pnpm fetch
 COPY . ./
 
 # install monorepo dependencies, from the virtual store
-RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --offline
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --prefer-offline
 
 # build all packages/apps
 RUN pnpm nx run-many --target=build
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -51,3 +51,26 @@ ENV NODE_EXTRA_CA_CERTS=/usr/local/share/ca-certificates/extra-ca-certificates.c
 EXPOSE 7310
 
 CMD [ "pnpm", "start" ]
+
+########################################
+# STEP 3: EVENT-LOG-INDEXER
+########################################
+
+FROM node:18.19-bullseye-slim AS event-log-indexer
```
nitaliano, line 59: Ponder uses `ast-grep` under the hood. When using the alpine image the app would crash at runtime because it couldn't find the expected architecture. For example https://github.com/ast-grep/ast-grep/tree/main/npm/platforms it might be just a bug with that version in Ponder or ast needs an update. Will look into this deeper at a later date.

jakim929 (approved):
looks great! the db setup script is gonna be really handy for testing permission limitations locally.

it looks like ponder resolved the issue regarding ast-grep so i think we can move to using the same image for all steps in the build. def doesn't have to be in this PR!

Issue #103, opened 2024-02-06
[m] Set up mixpanel in dapp console app


imranjami:
Implement a reusable way for us to have mixpanel tracking in applications in the ecosystem repo

tarunkhasnavis:
Update - we will only be adding Mixpanel at the app level starting with Dapp Console

PR #136, opened 2024-02-27
dapp console: add mixpanel 

tarunkhasnavis:
closes https://github.com/ethereum-optimism/ecosystem/issues/103

This is the initial PR for Mixpanel setup for Dapp Console. You can change mixpanel project tokens based on environment type (we have two: testing and production).


tarunkhasnavis:
* **#136** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/136?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/136?utm_source=stack-comment).

apps/dapp-console/app/mixpanel/mixpanel.tsx
```diff
@@ -0,0 +1,14 @@
+import mixpanelBrowser from 'mixpanel-browser'
+
+const initMixpanel = () => {
+  if (
```
nitaliano, line 4: would just do `if(!process.env.VITE_MIXPANEL_TOKEN)` then we can have the system setup so if it doesn't exist we just don't init mixpanel, we can ignore passing in incorrect strings

apps/dapp-console/app/mixpanel/mixpanel.tsx
```diff
@@ -0,0 +1,14 @@
...
+    process.env.VITE_MIXPANEL_TOKEN === 'false' ||
+    process.env.VITE_MIXPANEL_TOKEN == null
+  ) {
+    return null
+  }
+  mixpanelBrowser.init(process.env.VITE_MIXPANEL_TOKEN)
+  return mixpanelBrowser
+}
+
+export const mixpanel = initMixpanel()
```
nitaliano, line 14: not sure how you're going to implement the tracking portion, but i think we should avoid using this directly in our component files like we do in gateway meaning this

```
mixpanel?.track(TRACKING_EVENT_NAME.PRETEND_EVENT, 
  [MIXPANEL_CUSTOM_TRACKING_PROPERTY.PRETEND]: 'pretend',
})
```

having to import the custom tracking properties especially if we ever want to use the same event track call in multiple places and having mixpanel across the codebase would make it harder to refactor if we ever dropped mixpanel

instead i think we should wrap all them with functions and keep them under this directory so we can have functions that do this

```
function trackPretendEvent(args: PretendEventArgs) {
    mixpanel?.track('pretendEvent', 
     'Pretend': args.pretendProperty,
   })
}
```

nitaliano:
potentially also drop the `mixpanel` directory name for something more generic like `event-tracking` or something along those lines so the imports also stay generic

apps/dapp-console/app/mixpanel/mixpanel.tsx
tarunkhasnavis: oh cool, so the usage would be like 

```
import { trackPretendEvent } from 'event-tracking/mixpanel'

// on the action to track 
trackPretendEvent(relevantArgs)
```

and for each new event name there would be a new event function added to mixpanel.ts taking in the relevant arguments

nitaliano: Yup exactly and if the file ever gets too long can be broken out into multiple

nitaliano (approved):

PR #138, opened 2024-02-28
Set up nx generator for creating new trpc api servers

tremarkley:
Adds a new nx generator to the ecosystem repo: `trpc-api-generator` that can be used to add a new trpc api instance to the repo.

In a follow up PR will be adding logger, metrics and cors to the generator.

### DIFF
--- a/packages/api-plugin/package.json
+++ b/packages/api-plugin/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "@eth-optimism/api-plugin",
+  "version": "0.0.1",
+  "scripts": {
+    "build": "tsup",
+    "clean": "rm -rf dist",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint -c ./.eslintrc.cjs --no-eslintrc \"**/*.{ts,tsx}\" && pnpm prettier --check \"**/*.{ts,tsx}\"",
+    "lint:fix": "eslint -c ./.eslintrc.cjs --no-eslintrc \"**/*.{ts,tsx}\" --fix --quiet && pnpm prettier \"**/*.{ts,tsx}\" --write --loglevel=warn"
+  },
+  "dependencies": {
+    "@nx/devkit": "17.1.3",
+    "tslib": "^2.3.0"
+  },
+  "type": "commonjs",
+  "main": "./src/index.js",
+  "typings": "./src/index.d.ts",
+  "generators": "./generators.json",
+  "devDependencies": {
+    "tsup": "^8.0.1",
+    "typescript": "^5.2.2"
+  }
+}
--- a/packages/api-plugin/src/generators/trpc-api-generator/files/README.md
+++ b/packages/api-plugin/src/generators/trpc-api-generator/files/README.md
@@ -1,0 +1,17 @@
+# Service
+
+A barebones trpc server.
+
+## Getting started
+
+#### 1. Create `.env` file and fill in empty values
+
+```bash
+cp example.env .env
+```
+
+#### 2. Start dev server
+
+```bash
+pnpm run dev
+```
--- a/packages/api-plugin/src/generators/trpc-api-generator/files/src/Service.ts
+++ b/packages/api-plugin/src/generators/trpc-api-generator/files/src/Service.ts
@@ -1,0 +1,120 @@
+import bodyParser from 'body-parser'
+import type { Router } from 'express'
+import express from 'express'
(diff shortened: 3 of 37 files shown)


tremarkley:
* **#138** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/138?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/138?utm_source=stack-comment).

packages/api-plugin/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "api-plugin",
```
nitaliano, line 2: i think we should always prefix every package in the repo with `@eth-optimism/`

.github/actions/setup/action.yml
```diff
@@ -26,5 +26,5 @@ runs:
       shell: bash
 
     - name: pnpm build
-      run: pnpm nx affected --base=$NX_BASE --head=$NX_HEAD --target=build
+      run: pnpm nx affected --base=$NX_BASE --head=$NX_HEAD --target=build --exclude=api-plugin
```
nitaliano, line 29: Theres no `build` command for `api-plugin` so i don't think we need to worry about excluding `api-plugin` i think it would be good to make sure that all the linting atleast works for this. Since this is using nx affected it will only run this on changes so it won't build every run

packages/api-plugin/src/generators/trpc-api-generator/files/src/Service.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  private readonly routes = async (router: Router) => {
+    // user facing
+    router.use(
+      this.apiServerV0.expressRoute,
+      this.apiServerV0.createExpressMiddleware(),
+    )
+    router.use(
```
nitaliano, line 114: not something for this PR, but something for us to think about as a potential addon

it might be good to include the `ensureAdmin` middleware from gateway or something along those lines just to make sure we never accidentally deploy anything without realizing the playground is public for everyone to hit

packages/api-plugin/src/generators/trpc-api-generator/files/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "include": ["src"],
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./src/*"]
+    },
+    "target": "es2017",
```
nitaliano, line 9: i'm wondering if we should update this to be something more recent we have options for `2020` `2021` `2022` and `esnext`?

packages/api-plugin/src/generators/trpc-api-generator/files/tsup.config.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-console */
+import { defineConfig } from 'tsup'
+
+export default defineConfig({
```
nitaliano, line 4: it might be good to include `dts: true` in this so it will output `*.d.ts` files

packages/api-plugin/src/generators/trpc-api-generator/files/src/Service.ts
tremarkley: agreed!

packages/api-plugin/src/generators/trpc-api-generator/files/tsconfig.json
tremarkley: good call, might as well use the latest version we can

packages/api-plugin/src/generators/trpc-api-generator/files/tsup.config.ts
tremarkley: good point

nitaliano (approved):

packages/api-plugin/src/generators/trpc-api-generator/files/README.md
```diff
@@ -0,0 +1,17 @@
+# Paymaster Proxy Service
```
nitaliano, line 1: this should be updated in a follow on to remove the `Paymaster` name above and below

PR #159, opened 2024-03-07
[dapp console] Adds Feature Flag Support

nitaliano:
**Description**

Adds feature flag support to dapp console. These PR just wraps growthbook and ports some of our gateway code into the dapp console.

There are two new env vars
* `NEXT_PUBLIC_GROWTHBOOK_ENDPOINT`
* `NEXT_PUBLIC_GROWTHBOOK_ENCRYPTION_KEY` (only needed for the prod endpoint)

when it comes to configuring what endpoint we hit i'm going to configure it like so in netlify

**Production**: prod endpoint, encrypted (honestly not sure why it even needs to be encrypted but thats how the endpoint is setup)
**Deploy Previews**: staging endpoint, not encrypted
**Branch Deploys**: staging endpoint, not encrypted

if anyone wants to use this locally you can grab the values in netlify

### DIFF
--- a/apps/dapp-console/.eslintrc.cjs
+++ b/apps/dapp-console/.eslintrc.cjs
@@ -25,5 +25,6 @@ module.exports = {
       { varsIgnorePattern: '_', argsIgnorePattern: '_' },
     ],
     'react-refresh/only-export-components': ['off'],
+    'no-extra-semi': 'off'
   },
 }
--- a/apps/dapp-console/app/constants/envVars.ts
+++ b/apps/dapp-console/app/constants/envVars.ts
@@ -1,0 +1,18 @@
+import { z } from 'zod'
+
+const schema = z.object({
+  GROWTHBOOK_ENDPOINT: z
+    .string()
+    .url()
+    .default('https://cdn.growthbook.io/api/features/')
+    .describe('Growthbook endpoint'),
+  GROWTHBOOK_ENCRYPTION_KEY: z
+    .string()
+    .optional()
+    .describe('Key for decrypting growthbook response'),
+})
+
+export const ENV_VARS = schema.parse({
+  GROWTHBOOK_ENDPOINT: process.env.NEXT_PUBLIC_GROWTHBOOK_ENDPOINT,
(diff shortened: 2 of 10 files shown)


apps/dapp-console/package.json
```diff
@@ -4,14 +4,15 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "next build --no-lint",
```
nitaliano, line 7: this prevents next from using it's own linting rules and ignoring the ones we have setup

apps/dapp-console/app/constants/envVars.ts
```diff
@@ -0,0 +1,18 @@
...
+    .url()
+    .default('https://cdn.growthbook.io/api/features/')
+    .describe('Growthbook endpoint'),
+  GROWTHBOOK_ENCRYPTION_KEY: z
+    .string()
+    .optional()
+    .describe('Key for decrypting growthbook response'),
+})
+
+export const ENV_VARS = schema.parse({
```
nitaliano, line 15: we won't be able to pass `process.env` into a function like we do in gateway to generate the values at runtime, we need to do it statically at buildtime with next

apps/dapp-console/.eslintrc.cjs
```diff
@@ -25,5 +25,6 @@ module.exports = {
       { varsIgnorePattern: '_', argsIgnorePattern: '_' },
     ],
     'react-refresh/only-export-components': ['off'],
+    'no-extra-semi': 'off'
```
nitaliano, line 28: prettier adds a `;` to the beginning of immediately invoking functions which is the correct behavior to avoid ASI failure, and eslint has a silly rule `no-extra-semi` which is being brought in so the two are conflicting, turning this off for eslint

jakim929 (approved):

apps/dapp-console/.eslintrc.cjs
tremarkley: should this be included in all our eslintrc.cjs files?

PR #160, opened 2024-03-07
Configure docker for dapp console api

tremarkley:
Part of https://github.com/ethereum-optimism/ecopod/issues/870

Adds docker-compose for dapp-console-api.


tremarkley:
* **#160** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dockerfile
```diff
@@ -74,3 +74,25 @@ ENV NODE_EXTRA_CA_CERTS=/usr/local/share/ca-certificates/extra-ca-certificates.c
 EXPOSE 42069
 
 CMD [ "pnpm", "start" ]
+
+########################################
+# STEP 4: DAPP-CONSOLE-API
+########################################
+
+FROM node:18.19-bullseye-slim AS dapp-console-api
```
nitaliano, line 82: we'll want this docker image to look more like what the paymaster is doing https://github.com/ethereum-optimism/ecosystem/blob/7c10c041d22e1acf6247ffc3836af2d979fc9191/Dockerfile#L38-L53. we want the base OS to be alpine and not bullseyes, the reason the event-log-indexer is doing it this way is ponder needed a slightly different setup with more deps

jakim929: +1,

also one thing to note is the builder step builds **all** of the apps, so no need to rebuild, or refetch any dependencies

pnpm deploy the app then simply copy it in
```
RUN pnpm deploy --filter=dapp-console-api --prod /dapp-console-api

```

tremarkley: thank you, good suggestions!

jakim929 (approved):

apps/dapp-console-api/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      target: dapp-console-api
+    env_file: .env
+    environment:
+      - REDIS_URL=redis://redis:6379
+    healthcheck:
+      test: curl http://0.0.0.0:7300/api/v0/healthz
+    volumes:
+      - ../../certs/extra-ca-certificates.crt:/usr/local/share/ca-certificates/extra-ca-certificates.crt
+    ports:
+      - 7310:7300
```
jakim929, line 30: nit for later: wdyt about making this port something other than 7310 so it doesn't clash with paymaster-proxy

tremarkley: I can update that now!

PR #179, opened 2024-03-12
Adds Unlink Wallet Flow

nitaliano:
**Description**

Adds Unlink Wallet Flow

Screenshot 2024-03-13 at 11 53 49 AM
Screenshot 2024-03-13 at 11 53 40 AM

### DIFF
--- a/apps/dapp-console/app/settings/components/LinkedWallet.tsx
+++ b/apps/dapp-console/app/settings/components/LinkedWallet.tsx
@@ -30,7 +30,12 @@ export const LinkedWallet = ({ wallet, onUnlink }: LinkedWalletProps) => {
       <Input className="rounded" value={wallet.address} disabled />
       <Dialog open={isDialogOpen} onOpenChange={setIsDialogOpen}>
         <DialogTrigger asChild>
-          <Button className="rounded" variant="secondary" size="icon">
+          <Button
+            className="rounded"
+            variant="secondary"
+            size="icon"
+            aria-label="Unlink Wallet"
+          >
             <RiCloseLine />
           </Button>
         </DialogTrigger>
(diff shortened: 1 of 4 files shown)


apps/dapp-console/app/settings/components/LinkedWallet.tsx
```diff
@@ -0,0 +1,68 @@
...
+    () => setIsDialogOpen(false),
+    [setIsDialogOpen],
+  )
+
+  return (
+    <div className="flex flex-row gap-2">
+      <Input className="rounded" value={wallet.address} disabled />
+      <Dialog open={isDialogOpen} onOpenChange={setIsDialogOpen}>
+        <DialogTrigger asChild>
+          <Button className="rounded" variant="secondary" size="icon">
```
tarunkhasnavis, line 33: q: do we need `aria-label` when button text is not clear (maybe it is clear inside <RiCloseLine>)

apps/dapp-console/app/settings/components/LinkedWallet.tsx
```diff
@@ -0,0 +1,68 @@
...
+              Are you sure you want to unlink this wallet?
+            </Text>
+            <Text
+              as="p"
+              className="cursor-default mb-4 text-muted-foreground text-base"
+            >
+              {shortenAddress(wallet.address)}
+            </Text>
+
+            <Button className="rounded w-full mb-2" onClick={handleUnlink}>
```
tarunkhasnavis, line 53: Could be cool in the future to show toasts for linking and unlinking

tarunkhasnavis (approved):
lgtm! 

Whenever we add new packages should we be pushing the `pnpm-lock.yaml` file up?

apps/dapp-console/app/settings/components/LinkedWallet.tsx
nitaliano: yea i think semantically you do want to use an `aria-label` when a button just contains svg, will add!

PR #180, opened 2024-03-12
(dapp-console) set up postgres database for dapp-console-api

tremarkley:
Part of https://github.com/ethereum-optimism/ecopod/issues/872

This PR configures the dapp console api to connect to a postgres database. This also includes adding a `postgres` service to `docker-compose` so that the database can easily be spun up for local development.


tremarkley:
* **#180** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/180?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/dapp-console-api/docker-compose.yml
```diff
@@ -22,6 +33,8 @@ services:
     env_file: .env
     environment:
       - REDIS_URL=redis://redis:6379
+      - DB_HOST=postgres
+      - DB_PORT=5432
```
nitaliano, line 37: it might be worth setting all of the db env vars here instead of in the env file

apps/dapp-console-api/docker-compose.yml
```diff
@@ -22,6 +33,8 @@ services:
     env_file: .env
     environment:
       - REDIS_URL=redis://redis:6379
+      - DB_HOST=postgres
```
nitaliano, line 36: would set this to `dapp-console-api@oplabs-local-web.iam`, if you run into any connection errors from a fresh db run you'll most likely need to encode the `@`

apps/dapp-console-api/package.json
```diff
@@ -14,12 +14,16 @@
     "kill": "npx kill-port 7300",
     "build": "tsup && pnpm tsc",
     "start": "node build/cmd/run.js",
-    "docker:services:up": "docker compose up -d redis",
+    "docker:services:up": "docker compose up -d postgres redis",
```
nitaliano, line 17: would also add a docker compose down command here to make developing easier so we can tear it down faster

apps/dapp-console-api/docker-compose.yml
tremarkley: I think this needs to be set to `postgres` it's so the connection string points to the docker `postgres` service.

nitaliano (approved):
nice, good job!

apps/dapp-console-api/package.json
tremarkley: there is already a docker:down command. are you saying there should be another one that just brings down the services?

nitaliano: it's probably fine to use the other one, what i'm wondering will happen is if someone is developing thats not as used to the backend code running the down command will bring down the db and wipe it which might be unexpected if you're thinking it's only bringing down the api server

the way the other one is setup allows you to pick what you want to bring down which is great and allows us to pick and choose and probably good for the rest of the team to continue to get used to compose more

apps/dapp-console-api/docker-compose.yml
tremarkley: yeah, I like that.

PR #183, opened 2024-03-14
feat(dapp-console): create entities data model

tremarkley:

PR #194, opened 2024-03-20
feat(dapp-console): create address data model

tremarkley:
Closes https://github.com/ethereum-optimism/ecopod/issues/875

### DIFF
--- a/apps/dapp-console-api/src/models/addresses.ts
+++ b/apps/dapp-console-api/src/models/addresses.ts
@@ -1,0 +1,49 @@
+import {
+  index,
+  jsonb,
+  pgTable,
+  timestamp,
+  uniqueIndex,
+  uuid,
+  varchar,
+} from 'drizzle-orm/pg-core'
+import type { Address } from 'viem'
+
+import { entities } from './entities'
+
+enum AddressState {
+  ACTIVE = 'active',
+  DISABLED = 'disabled',
+  SANCTIONED = 'sanctioned',
+}
+
+type AddressVerification = {}
+
+export const addresses = pgTable(
+  'addresses',
+  {
+    createdAt: timestamp('created_at', { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+    updatedAt: timestamp('updated_at', { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+    id: uuid('id').defaultRandom().primaryKey(),
+    entityId: uuid('entity_id').references(() => entities.id),
+    address: varchar('address').$type<Address>().notNull(),
+    verifications: jsonb('verifications')
+      .$type<AddressVerification>()
+      .default({})
+      .notNull(),
+    state: varchar('state')
+      .$type<AddressState>()
+      .default(AddressState.ACTIVE)
+      .notNull(),
+  },
+  (table) => {
+    return {
+      entitiyIdAddressIdx: uniqueIndex().on(table.entityId, table.address),
+      addressIdx: index().on(table.address),
+    }
+  },
+)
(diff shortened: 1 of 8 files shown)


tremarkley:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/194?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#194** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/194?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#183** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/183?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/dapp-console-api/src/models/addresses.ts
```diff
@@ -0,0 +1,43 @@
...
+    verifications: jsonb('verifications').default({}).notNull(),
+    state: varchar('state')
+      .$type<AddressState>()
+      .default(AddressState.ACTIVE)
+      .notNull(),
+  },
+  (table) => {
+    return {
+      entitiyIdAddressIdx: uniqueIndex().on(table.entityId, table.address),
+      addressIdx: index('address_idx').on(table.address),
```
nitaliano, line 40: We should avoid passing in strings to the index functions with drizzle. Index names exist on the global namespace across the database so this will create an index called `address_idx` which will conflict with other tables if they attempt to call it that. When we don't pass a string into the function drizzle will namespace it with the table name automatically so it won't collide.

apps/dapp-console-api/migrations/0001_broad_drax.sql
```diff
@@ -0,0 +1,17 @@
...
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"entity_id" uuid,
+	"address" varchar NOT NULL,
+	"verifications" jsonb DEFAULT '{}'::jsonb NOT NULL,
+	"state" varchar DEFAULT 'active' NOT NULL
+);
+--> statement-breakpoint
+CREATE UNIQUE INDEX IF NOT EXISTS "addresses_entity_id_address_index" ON "addresses" ("entity_id","address");--> statement-breakpoint
```
nitaliano, line 11: this is what i mean in my other command you can see how it's namespaced compared to below.

apps/dapp-console-api/src/models/addresses.ts
tremarkley: makes sense, good catch

nitaliano (approved):

apps/dapp-console-api/src/models/addresses.ts
```diff
@@ -0,0 +1,49 @@
...
+  'addresses',
+  {
+    createdAt: timestamp('created_at', { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+    updatedAt: timestamp('updated_at', { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+    id: uuid('id').defaultRandom().primaryKey(),
+    entityId: uuid('entity_id').references(() => entities.id),
```
nitaliano, line 32: realized this after looking at the other PR, which should make this not null we don't want to be able to save anything to the db without an entity ref

tremarkley: good catch!

PR #213, opened 2024-03-21
Feat: op-app styles

kelreel:
**Description**

Simple styling and small refactoring of the `op-app`.

Fixed Invalid chain config (wagmi) on page refresh.

image


nitaliano (approved):
Nice changes, thanks for another PR 😄! We might also turn this test app into a playground for viem and wagmi in the future.

PR #218, opened 2024-03-22
[dapp console] Adds copy address and toast support

nitaliano:
**Description**

* Adds the `Toaster` component to our root layout. This allows us to start showing toasts in the app https://ui.shadcn.com/docs/components/toast
* Updates the wallets page to add the copy icon, and display a toast after copying

https://github.com/ethereum-optimism/ecosystem/assets/1761993/1fcea4e5-6348-4330-8563-9ca496d5f752

### DIFF
--- a/apps/dapp-console/app/settings/components/LinkedWallet.tsx
+++ b/apps/dapp-console/app/settings/components/LinkedWallet.tsx
@@ -8,8 +8,9 @@ import { Input } from '@eth-optimism/ui-components/src/components/ui/input/input
 import { Button } from '@eth-optimism/ui-components/src/components/ui/button/button'
 import { Text } from '@eth-optimism/ui-components/src/components/ui/text/text'
 import { WalletWithMetadata } from '@privy-io/react-auth'
-import { RiAlertFill, RiCloseLine } from '@remixicon/react'
+import { RiAlertFill, RiCloseLine, RiFileCopyLine } from '@remixicon/react'
 import { useCallback, useState } from 'react'
+import { useToast } from '@eth-optimism/ui-components'
 
 export type LinkedWalletProps = {
   wallet: WalletWithMetadata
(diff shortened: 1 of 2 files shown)


apps/dapp-console/app/settings/components/LinkedWallet.tsx
```diff
@@ -25,12 +27,34 @@ export const LinkedWallet = ({ wallet, onUnlink }: LinkedWalletProps) => {
     [setIsDialogOpen],
   )
 
+  const handleCopy = useCallback(() => {
+    navigator.clipboard.writeText(wallet.address)
+    toast({
+      description: 'Address Copied',
+      duration: 1000,
```
nitaliano, line 34: we should eventually have these as common values somewhere

apps/dapp-console/app/settings/components/LinkedWallet.tsx
```diff
@@ -25,12 +27,34 @@ export const LinkedWallet = ({ wallet, onUnlink }: LinkedWalletProps) => {
...
+    })
+  }, [wallet, toast])
+
   return (
     <div className="flex flex-row gap-2">
       <Input value={wallet.address} disabled />
+      <Button
+        variant="secondary"
+        className="px-3"
```
tremarkley, line 43: definitely dont do in this pr and should be a broader discussion with the team. but I wonder if we should make these tailwind classes constants, so if we ever do want to upgrade them across all of our design system, it is an easy change. cc @imranjami @tarunkhasnavis @jakim929

tremarkley (approved):

apps/dapp-console/app/settings/components/LinkedWallet.tsx
nitaliano: i'm not convinced we actually will need to add padding, i was just seeing if it will look better or not without adding padding right now i think the buttons look like eggs

Screenshot 2024-03-22 at 1 11 22 PM

PR #224, opened 2024-03-26
feat(api-key-service): create new app

jakim929:
creates new trpc api service for the api-key-service

mostly uses the trpc-api generator but changes up how we do dependency injection for metrics / logger since we no longer have to do hacky "setLoggingServer"

if changes look good i can change the generator files as well

### DIFF
--- a/apps/api-key-service/src/routes/Route.ts
+++ b/apps/api-key-service/src/routes/Route.ts
@@ -1,0 +1,39 @@
+import type { BuildProcedure, CreateRouterInner } from '@trpc/server'
+import type { Logger } from 'pino'
+import { z } from 'zod'
+
+import type { Metrics } from '@/monitoring/metrics'
+
+import type { Trpc } from '../Trpc'
+
+/**
+ * Class routes extend to create a TRPC route.
+ * Every route is a name and a controller.
+ * Zod is used to create validation
+ */
+export abstract class Route {
+  /**
+   * quality of life property.  Zod is used to validate input in a
+   * typesafe way.
+   * Can be used by `import {z} from 'zod' as well.
+   * @see https://github.com/colinhacks/zod
+   */
+  protected readonly z = z
+  /**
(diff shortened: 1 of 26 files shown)


jakim929:
* **#224** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/224?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/api-key-service/src/routes/Route.ts
```diff
@@ -0,0 +1,39 @@
...
+   * The handler specifies the typescript types for input and output
+   * and writes the function to fetch the data
+   */
+  public abstract readonly handler:
+    | BuildProcedure<any, any, any>
+    | CreateRouterInner<any, any>
+  constructor(
+    protected readonly trpc: Trpc,
+    protected readonly logger: Logger,
+    protected readonly metrics: Metrics,
```
jakim929, line 37: main difference from the current generator - injects loggers / metrics without need for setLoggingServer

apps/api-key-service/src/constants/session.ts
```diff
@@ -0,0 +1,23 @@
+import { type IronSession, type SessionOptions } from 'iron-session'
+
+import { envVars } from './envVars'
+
+export const sessionOptions: SessionOptions = {
+  password: envVars.IRON_SESSION_SECRET,
+  cookieName: `api-key-service.${envVars.DEPLOYMENT_ENV}.auth.session-token`,
```
nitaliano, line 7: we can probably remove all the cookie related logic, if we need anything on the admin api side of things we'll just proxy it thru the dapp console api to this service

nitaliano (approved):

apps/api-key-service/src/constants/session.ts
jakim929: updated to remove all admin api related code

PR #227, opened 2024-03-27
feat(dapp-console): add endpoint for syncing privy addresses to db

tremarkley:
Part of https://github.com/ethereum-optimism/ecopod/issues/890

Adds an endpoint to `/wallets` that fetches the wallets associated with a privy did and updates the db to contain those addresses. This also handles unlinking addresses that are no longer linked with the privy account.


tremarkley:
* **#227** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/227?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/dapp-console-api/src/models/wallets.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+
+enum LinkType {
+  PRIVY = 'privy',
+}
+
+type AddressVerification = {}
+
+export const wallets = pgTable(
+  'wallets',
```
tremarkley, line 33: @nitaliano since I believe this table will primarily be used for dealing with the accounts associated with a user I decided to rename it to wallets. I also added a `linkType` column for determining what provider was used for linking the wallet. I imagine in the future we may have other did providers or even the option for users to manually link their wallets. Let me know what you think of this change.

apps/dapp-console-api/src/testhelpers/auth.ts
```diff
@@ -1,21 +1,16 @@
...
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { Mock } from 'vitest'
 
 import type { SessionData } from '@/constants'
-import { PRIVY_TOKEN_COOKIE_KEY } from '@/constants'
+import { envVars, PRIVY_TOKEN_COOKIE_KEY } from '@/constants'
 import type { Session } from '@/constants/session'
-import { getEntity, insertEntity } from '@/models'
 
-vi.mock('../models', async () => ({
```
tremarkley, line 13: this was causing a bunch of issues with trying to use the helpers in this file while also having a different set of mocks, so removing entirely.

apps/dapp-console-api/src/testhelpers/auth.ts
```diff
@@ -49,17 +44,17 @@ export const createSignedOutCaller = <T extends AnyRouter>(router: T) => {
   })
 }
 
-/** Sets up the mocks required for authentication. */
-export const configureAuthMocks = (privyClient: PrivyClient) => {
```
tremarkley, line 53: after re-thinking this, I realized there is no clean way to share these mocks and instead should export a session that will allow a user to pass the `isPrivyAuthed` middleware instead.

apps/dapp-console-api/src/testhelpers/auth.ts
```diff
@@ -1,21 +1,16 @@
-import type { PrivyClient } from '@privy-io/server-auth'
 import type { AnyRouter } from '@trpc/server'
 import { createCallerFactory } from '@trpc/server'
+import bcrypt from 'bcrypt'
 import type { Request, Response } from 'express'
 import type { getIronSession } from 'iron-session'
-import { type Mock, vi } from 'vitest'
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
tremarkley, line 6: this prevents the following type error: 
Screenshot 2024-03-28 at 3 50 35 PM

apps/dapp-console-api/src/testhelpers/session.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
tremarkley, line 1: this prevents the following type error:
Screenshot 2024-03-28 at 3 51 15 PM

apps/dapp-console-api/src/routes/wallets/WalletsRoute.ts
```diff
@@ -0,0 +1,43 @@
...
+      try {
+        const { user } = ctx.session
+
+        await fetchAndSyncPrivyWallets(
+          this.trpc.database,
+          this.trpc.privy,
+          user!.privyDid,
+          user!.entityId,
+        )
+
```
nitaliano, line 24: one thing i think we'll want to add to this endpoint in a follow on is to loop thru all the newly added wallets and do the verification checks

apps/dapp-console-api/src/utils/privy.ts
```diff
@@ -0,0 +1,120 @@
...
+) => {
+  return Promise.all(
+    entityWallets.map((entityWallet) => {
+      if (entityWallet.linkType !== WalletLinkType.PRIVY) {
+        return Promise.resolve({})
+      }
+
+      if (
+        entityWallet.state === WalletState.ACTIVE &&
+        !privyAddresses.includes(entityWallet.address)
```
nitaliano, line 99: I don't think we'll encounter this when checking addresses because it's coming from privy. Generally addresses can appear in two forms as a checksummed versions and a non checksummed version. They are the same address but the non checksummed version would be all lowercased, while the checksummed version would also include upper case letters. When comparing addresses it might be good to get in the habit of checking them by always converting them to lowercase

nitaliano (approved):
looks good!

apps/dapp-console-api/src/utils/privy.ts
tremarkley: good catch!

apps/dapp-console-api/src/routes/wallets/WalletsRoute.ts
tremarkley: yeah, that is a good idea

PR #230, opened 2024-03-28
feat(api-key-service): add api-keys table

jakim929:
adds a `api-keys` table to the api-key-service db
- not running any runtime migrations right now, will add logic to do this once deployment is set up

```
    id: uuid('id').defaultRandom().primaryKey(),
    entityId: uuid('entity_id').notNull(),
    key: varchar('key').notNull(),
    state: varchar('state').$type<ApiKeyState>().notNull(),
    stateUpdatedAt: timestamp('state_updated_at', { withTimezone: true }),
    deletedAt: timestamp('deleted_at', { withTimezone: true }),
    createdAt: timestamp('created_at', { withTimezone: true })
      .defaultNow()
      .notNull(),
    updatedAt: timestamp('updated_at', { withTimezone: true })
      .defaultNow()
      .notNull(),
```

### DIFF
--- a/apps/api-key-service/src/models/apiKeys.ts
+++ b/apps/api-key-service/src/models/apiKeys.ts
@@ -1,0 +1,35 @@
+import {
+  index,
+  pgTable,
+  timestamp,
+  uniqueIndex,
+  uuid,
+  varchar,
+} from 'drizzle-orm/pg-core'
+
+export type ApiKeyState = 'enabled' | 'disabled'
+
+export const apiKeys = pgTable(
+  'api-keys',
+  {
+    id: uuid('id').defaultRandom().primaryKey(),
+    entityId: uuid('entity_id').notNull(),
+    key: varchar('key').notNull(),
+    state: varchar('state').$type<ApiKeyState>().notNull(),
+    stateUpdatedAt: timestamp('state_updated_at', { withTimezone: true }),
+    deletedAt: timestamp('deleted_at', { withTimezone: true }),
+    createdAt: timestamp('created_at', { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+    updatedAt: timestamp('updated_at', { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+  },
+  (table) => {
+    return {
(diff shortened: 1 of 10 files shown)


jakim929:
* **#230** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/230?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/api-key-service/src/models/apiKeys.ts
```diff
@@ -0,0 +1,45 @@
+import {
+  index,
+  pgEnum,
+  pgTable,
+  timestamp,
+  uuid,
+  varchar,
+} from 'drizzle-orm/pg-core'
+
+export const apiKeyStateEnum = pgEnum('api_key_state', [
```
nitaliano, line 10: We were planning on using strings instead of enums, we actually haven't granted the permission for enums in our script. Enums don't get stored in the database as a plain string, so if we ever imported our data from the db into another system for data science or audit purposes they would also need to import some mapping of strings to figure out what means what. In generally i think using just strings and enforcing with typechecks on the codelevel will be better in the long term

jakim929: got it - will replace in favor of strings typechecked at app level

apps/api-key-service/src/models/apiKeys.ts
```diff
@@ -0,0 +1,28 @@
+import { index, pgTable, timestamp, uuid, varchar } from 'drizzle-orm/pg-core'
+
+export type ApiKeyState = 'enabled' | 'disabled' | 'deleted'
+
+export const apiKeys = pgTable(
+  'api-keys',
+  {
+    id: uuid('id').defaultRandom().primaryKey(),
+    entityId: uuid('entity_id').notNull(),
```
jakim929, line 9: this should point to the entityId in the dapp-console-api. 

not meant to be used for restrict read/write at the api-key-service level, more here for bookkeeping purposes, and for indexing api keys by creator

nitaliano (approved):

PR #240, opened 2024-04-05
Example One Click Deployer

nitaliano:
**Description**

This is a very slimmed down version of what the MVP for a 1 click deployer could look like and it's just built against anvil for now. This just shows how we can connect a github action to our dapp console api that can communicate with a Create2Factory deployer that we have onchain that will allow us to deploy any contract just based on it's bytecode and initialization arguments.

**Example Action Interface for an ERC20**
```
name: Deploy Contracts

jobs:
  deploy-erc20:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Set up foundry
        uses: foundry-rs/foundry-toolchain@v1
      - name: Build Contract Artifacts
        run: pnpm nx run @eth-optimism/contracts-ecosystem:compile
        shell: bash
      - name: Deploy Simple ERC20
        uses: ./apps/dapp-console-api/src/actions   
        with:
          args: Example,exa,18
          target: ./packages/contracts-ecosystem/src/SimpleERC20.sol:SimpleERC20
          buildDir: ./packages/contracts-ecosystem/out
          version: 1.0
```

**MVP using a database to store deployment info**
Screenshot 2024-04-05 at 3 21 00 PM

**MVP using a EAS to store deployment info**
Screenshot 2024-04-05 at 3 23 12 PM

**Upgradability**
This PR doesn't have anything regarding upgradability. I did explore what we need to do and will add more about this in our scoping document. All we'll need to do is update the `Create2Factory` to also deploy a proxy contract so we can always add a new implemention contract to it. Generally I think we would use an already existing [ERC1967Factory](https://eips.ethereum.org/EIPS/eip-1967) that would allow us to deploy a proxy contract alongside the initial implementation contract. This will require contracts that want upgradability to not have a constructor and instead have an `initalize` function that we end up calling after deployment with the initialization arguments supplied in the action.


packages/contracts-ecosystem/src/Create2Factory.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.8.24;
+
+import {StdUtils} from "forge-std/StdUtils.sol";
+
+contract Create2Factory {
```
nitaliano, line 5: this is just for an example we should never actually use this contract without adding real safety/auth logic to it

nitaliano:
Closing this out since it was part of our hackathon. We'll be using this as a reference going forward.

PR #245, opened 2024-04-09
feat(dapp-console-api): `listApps` endpoint

tremarkley:
Part of https://github.com/ethereum-optimism/ecopod/issues/925

Adds `listApps` endpoint for listing the active apps owned by an entity.

### DIFF
--- a/apps/dapp-console-api/src/routes/apps/AppsRoute.ts
+++ b/apps/dapp-console-api/src/routes/apps/AppsRoute.ts
@@ -1,0 +1,72 @@
+import { generateListResponse, zodListRequest, zodNameCursor } from '@/api'
+import { isPrivyAuthed } from '@/middleware'
+import { getActiveAppsForEntityByCursor, getContractsForApp } from '@/models'
+import { metrics } from '@/monitoring/metrics'
+import { Trpc } from '@/Trpc'
+
+import { DEFAULT_PAGE_LIMIT } from '../constants'
+import { Route } from '../Route'
+
+export class AppsRoute extends Route {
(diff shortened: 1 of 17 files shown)


tremarkley:
* **#245** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/245?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/dapp-console-api/src/routes/apps/AppsRoute.ts
```diff
@@ -0,0 +1,54 @@
...
+    .query(async ({ ctx, input }) => {
+      try {
+        const { user } = ctx.session
+        const limit = input.limit ?? DEFAULT_PAGE_LIMIT
+
+        if (!user) {
+          throw Trpc.handleStatus(401, 'user not authenticated')
+        }
+
+        const activeApps = await getActiveAppsForEntityByCursor({
```
tremarkley, line 29: for now, I think that this endpoint should only load the apps and we should have a separate endpoint for loading the contracts associated with an app. Other option would be to load the contracts alongside the apps and have this endpoint return a list of apps along with a list of contracts within those apps. I like having two separate endpoints with each one returning its own cursor since that will simplify things but downside is it requires an extra call from the client. wdyt?

tremarkley: ok, yeah, that definitely makes sense. updated to return contracts alongside each app.

nitaliano (approved):

PR #248, opened 2024-04-10
feat(dapp-console-api): add block timestamp column to transactions table

tremarkley:
Closes https://github.com/ethereum-optimism/ecopod/issues/917

### DIFF
--- a/apps/dapp-console-api/src/models/transactions.ts
+++ b/apps/dapp-console-api/src/models/transactions.ts
@@ -68,6 +68,8 @@ export const transactions = pgTable(
     value: bigint('value', { mode: 'bigint' }),
     /** `success` if this transaction was successful or `reverted` if it failed */
     status: varchar('status').$type<'success' | 'reverted'>().notNull(),
+    /** Unix timestamp of when this block was collated */
+    blockTimestamp: integer('block_timestamp').notNull(),
   },
   (table) => {
     return {
@@ -78,6 +80,7 @@ export const transactions = pgTable(
       fromAddressIdx: index().on(table.fromAddress),
       toAddressIdx: index().on(table.toAddress),
       contractAddressIdx: index().on(table.contractAddress),
+      entityBlockTimestampIdx: index().on(table.entityId, table.blockTimestamp),
     }
   },
 )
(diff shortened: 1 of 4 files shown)


tremarkley:
* **#248** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/248?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/dapp-console-api/src/models/transactions.ts
```diff
@@ -78,6 +80,7 @@ export const transactions = pgTable(
       fromAddressIdx: index().on(table.fromAddress),
       toAddressIdx: index().on(table.toAddress),
       contractAddressIdx: index().on(table.contractAddress),
+      blockTimestampIdx: index().on(table.blockTimestamp),
```
jakim929, line 83: curious what's a query pattern where this this index is necessary?

apps/dapp-console-api/src/models/transactions.ts
```diff
@@ -68,6 +68,8 @@ export const transactions = pgTable(
     value: bigint('value', { mode: 'bigint' }),
     /** `success` if this transaction was successful or `reverted` if it failed */
     status: varchar('status').$type<'success' | 'reverted'>().notNull(),
+    /** Unix timestamp of when this block was collated */
+    blockTimestamp: bigint('block_timestamp', { mode: 'bigint' }).notNull(),
```
jakim929, line 72: nit: block timestamps are all seconds since epoch so int should be fine I think

apps/dapp-console-api/src/models/transactions.ts
jakim929: might be better to index by blockNumber if the query pattern is look for all transactions from a specific block

jakim929 (approved):

apps/dapp-console-api/src/models/transactions.ts
tremarkley: I added this because the current figma designs have the ui sorting transactions by timestamp and also if we want to paginate these results we'll need a query pattern for querying transactions within a certain time range.

jakim929: is the design for sorting transactions for a single entity?

if the query is sort all transactions **for a given entity** by timestamp we prob want to add a multicolumn index on
(entityId, blockTimestamp). 

having them as two separate indices will not be as helpful with the sorting unless we're sorting the entire table by createdAt

PR #253, opened 2024-04-11
[dapp console] Integrate Wallets and Contracts pages with backend

nitaliano:
**Description**

**Error Handling**
* Adds a `captureError` function to make it easier to send errors to sentry with a consistent action to make debugging easier, so we know in what context the error happened
* Calls `setUser` from sentry after a user logins so all errors that get sent to sentry include the user id

**Wallets Page**
* Updates setting page to `syncWallets` after linking/unlinking

**Contracts Page**
* Fully integrates the contracts page with the backend for listing/creating apps, editing app names, creating contracts, and verifying them manually or automatically with a connected wallet.
* Adds `StartVerificationDialog` this is responsible for the whole entire verification flow. It comes with a provider called `ContractVerificationProvider` that implements a hook `useContractVerification` to keep all the different dialog content screens clean.
* Refactors how we're rendering apps and contracts.
* It's configured to just create contracts right now with op sepolia, this will be updated in a follow on PR once I add the chain dropdown

**Api Types**
* Adds a new `types/api.ts` file that has the type objects from the backend to make writing props easier

**Create Contract Route**
* Updates the endpoint to check if we have ever verified the deployer address before or not for the entity. If we have when we create the contract we mark it as verified


apps/dapp-console/app/settings/contracts/AddContractForm.tsx
```diff
@@ -39,20 +47,46 @@ const addContractSchema = z.object({
...
+  const handleCreateContract = useCallback(async () => {
+    try {
+      const formValues = form.getValues()
+
+      let contractToVerifiy = unverifiedContract
+
+      if (!contractToVerifiy) {
+        const { result } = await createContract({
+          appId,
+          chainId: optimismSepolia.id,
```
nitaliano, line 77: will implement the dropdown in the next PR since this is getting pretty big

apps/dapp-console-api/src/models/contracts.ts
```diff
@@ -106,6 +106,28 @@ export const getContract = async (input: {
   return results[0] || null
 }
 
+export const hasAlreadyVerifiedDeployer = async (input: {
```
nitaliano, line 109: will follow on with tests in the next PR

apps/dapp-console-api/src/routes/contracts/ContractsRoute.ts
```diff
@@ -177,7 +184,9 @@ export class ContractsRoute extends Route {
               deployerAddress,
               chainId,
               appId,
-              state: ContractState.NOT_VERIFIED,
+              state: isDeployerVerified
```
nitaliano, line 187: The FE reads the state of the contract after it's been created to decide if it needs to show the verification flow or just show a success screen

apps/dapp-console/app/settings/contracts/AddContractForm.tsx
```diff
@@ -118,7 +152,7 @@ export const AddContractForm = ({
       <div className="mt-3 flex flex-row justify-end w-full">
         <Button
           disabled={!form.formState.isValid}
```
tremarkley, line 154: nit: feel free to add in follow up pr but should there be a loading state on the button?

apps/dapp-console/app/settings/contracts/DeployedApp.tsx
```diff
@@ -0,0 +1,129 @@
...
+    [setAppName],
+  )
+
+  const handleStartVerification = useCallback(
+    async (contract: Contract) => {
+      setContractToVerify(contract)
+      setVerifyDialogOpen(true)
+
+      const { data: contracts } = await fetchContracts()
+      setContracts(contracts as Contract[])
```
tremarkley, line 54: how come this has to be casted?

tremarkley (approved):

apps/dapp-console/app/settings/contracts/AddContractForm.tsx
nitaliano: yup will do! i haven't added any loading states yet be will when i start to add them once the backend integration is done

apps/dapp-console/app/settings/contracts/DeployedApp.tsx
nitaliano: i think it's because of the enum/string difference

PR #257, opened 2024-04-11
feat(dapp-console-api): `startVerification` endpoint

tremarkley:
Part of https://github.com/ethereum-optimism/ecopod/issues/891

Adds `startVerification` endpoint to the `contracts` route. This endpoint handles creating a challenge for the user when they are going through the contract verification flow.

### DIFF
--- a/apps/dapp-console-api/src/routes/contracts/ContractsRoute.ts
+++ b/apps/dapp-console-api/src/routes/contracts/ContractsRoute.ts
@@ -5,9 +5,18 @@ import {
 } from '@/api'
 import { supportedChainsPublicClientsMap } from '@/constants'
 import { isPrivyAuthed } from '@/middleware'
-import { ContractState, getContractsForApp, insertContract } from '@/models'
+import {
+  ChallengeState,
+  ContractState,
+  getContract,
+  getContractsForApp,
+  getUnexpiredChallenge,
+  insertChallenge,
+  insertContract,
+} from '@/models'
 import { metrics } from '@/monitoring/metrics'
 import { Trpc } from '@/Trpc'
+import { generateChallenge } from '@/utils'
 
 import { Route } from '../Route'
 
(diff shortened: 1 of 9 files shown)


tremarkley:
* **#257** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/257?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/dapp-console-api/migrations/0012_conscious_khan.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "challenges" DROP COLUMN IF EXISTS "challenge";--> statement-breakpoint
+ALTER TABLE "challenges" DROP COLUMN IF EXISTS "reponse";
```
tremarkley, line 2: dropping the challenge column. I think it is best if we generate challenges on request instead of storing in db. this gives us the ability to change challenges later if we need to. 

dropping the response column because there is no need to store signatures for this feature.

apps/dapp-console-api/src/routes/contracts/ContractsRoute.ts
```diff
@@ -146,7 +156,102 @@ export class ContractsRoute extends Route {
...
+      return { ...result, challenge: challengeToComplete }
+    })
+
   public readonly handler = this.trpc.router({
     [this.listContractsForApp]: this.listContractsForAppController,
+    [this.createContract]: this.createContractController,
+    [this.startVerification]: this.startVerificationController,
   })
+
+  private generateChallenge = (address: Address) => {
```
nitaliano, line 254: it's probably fine for now but maybe we should have a util file for challenges or move this into the model

apps/dapp-console-api/migrations/0012_conscious_khan.sql
nitaliano: i think this is fine, something to consider is if we ever changed this while there were pending challenges out it could cause them to start failing if we're not keeping track of them or what version of a challenge is being signed. for an initial version of this though i think this is fine until we otherwise run into a need to keep it in the db

PR #281, opened 2024-04-19
Rebate Flows

nitaliano:
**Description**

* Adds a `HexBadge` component used for displaying any sort of hex with a label and the shortened version of the hex
* Adds a `NetworkBadge` component that takes in a chain id and will display the chain name and it's logo
* Adds a `Network` component this is a general purpose component used that displays the chain name and logo in a flex row, it's being used in other components that need network info
* Adds a `superchain.ts` file that contains all the superchain defs. We'll eventually move this to a common pkg.
* Updates contract creation flow to use relations returned in request
* Updates rebate flows to be integrated with backend except the claim mutation
* Update our api types to be inferred from the backend, still not crazy about how it's being inferred something we'll need to address so we can pull these types from the backend instead of inference for our primitive objects.

### DIFF
--- a/apps/dapp-console/app/components/Selects/L2NetworkSelect.tsx
+++ b/apps/dapp-console/app/components/Selects/L2NetworkSelect.tsx
@@ -1,0 +1,82 @@
+import { SuperchainNetwork, superchain } from '@/app/constants/superchain'
+import {
+  Select,
+  SelectContent,
+  SelectGroup,
+  SelectItem,
+  SelectTrigger,
+} from '@eth-optimism/ui-components'
+import { useCallback, useMemo, useState } from 'react'
+import { Chain } from 'viem'
+import { optimism } from 'viem/chains'
+import { Network } from '@/app/components/Network'
+
+export type SuperchainDropdownProps = {
+  includeTestnets?: boolean
+  onNetworkChange?: (network: L2NetworkSelectItem) => void
+}
+
+export type L2NetworkSelectItem = {
+  id: number
+  chain: Chain
+  config: SuperchainNetwork
+}
+
+export const L2NetworkSelect = ({
+  includeTestnets,
+  onNetworkChange,
+}: SuperchainDropdownProps) => {
+  const [selectedNetwork, setSelectedNetwork] = useState<L2NetworkSelectItem>({
+    id: optimism.id,
+    chain: optimism,
+    config: superchain.op,
+  })
+
+  const allChains = useMemo(() => {
+    return Object.values(superchain).reduce((arr, config) => {
+      if (config.mainnet) {
+        arr.push({ id: config.mainnet.id, chain: config.mainnet, config })
+      }
+
+      if (includeTestnets && config.testnets) {
+        for (const testnet of config.testnets) {
+          arr.push({ id: testnet.id, chain: testnet, config })
+        }
+      }
+
+      return arr
+    }, [] as L2NetworkSelectItem[])
+  }, [includeTestnets])
+
+  const handleNetworkChange = useCallback(
+    (chainIdStr: string) => {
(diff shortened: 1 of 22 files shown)


apps/dapp-console/app/settings/contracts/page.tsx
```diff
@@ -6,8 +6,17 @@ import { ClaimedRebateProgressBanner } from '@/app/settings/components/ClaimedRe
 import { CoinbaseVerificationBanner } from '@/app/settings/components/CoinbaseVerificationBanner'
 import { DeployedApps } from '@/app/settings/contracts/DeployedApps'
 import { ClaimedRebates } from '@/app/settings/components/ClaimedRebates'
+import { apiClient } from '@/app/helpers/apiClient'
+import { useEffect } from 'react'
 
 export default function Contracts() {
+  const { mutate: syncCbVerification } =
```
nitaliano, line 13: Adding this sync so it always happens on the contract page, so people can get unstuck with a reload. I think we should look into sharing the cb verified code so the FE can start polling this when we have the rebate dialog modal open. We can poll EAS from the FE and when we find an attestation we call this sync from the FE.

apps/dapp-console/app/settings/components/ClaimRebateDialog.tsx
```diff
@@ -8,39 +8,50 @@ import {
...
-  contractAddress,
-  contractDeploymentTransaction,
-  amountClaimed = BigInt(0),
-  isVerified = false,
+  contract,
   children,
 }: ClaimRebateDialogProps) => {
   const [isOpen, setOpen] = useState(false)
 
+  const { data: wallets } = apiClient.wallets.listWallets.useQuery({})
```
tremarkley, line 32: if you're just looking for wallet verifications, the `walletVerifications` endpoint might be better here: https://github.com/ethereum-optimism/ecosystem/blob/dfaba331691e610cdcbb4154b43778007709e17e/apps/dapp-console-api/src/routes/wallets/WalletsRoute.ts#L128

apps/dapp-console/app/settings/components/ClaimRebateDialog.tsx
```diff
@@ -8,39 +8,50 @@ import {
...
+  const { data: totalRebateAmount } =
+    apiClient.Rebates.totalRebatesClaimed.useQuery()
+  const amountClaimed = useMemo(
+    () => totalRebateAmount ?? BigInt(0),
+    [totalRebateAmount],
+  )
+
   const gasFeeToReimburst = useMemo(() => {
-    const gasAmount = contractDeploymentTransaction.gas
+    const gasAmount = BigInt(contract.transaction?.gasUsed ?? 0)
```
tremarkley, line 46: should this be gasUsed * gasPrice? that is how we calculate it on the backend: https://github.com/ethereum-optimism/ecosystem/blob/dfaba331691e610cdcbb4154b43778007709e17e/apps/dapp-console-api/src/routes/rebates/RebatesRoute.ts#L261

tremarkley (approved):

apps/dapp-console/app/components/Selects/L2NetworkSelect.tsx
```diff
@@ -0,0 +1,82 @@
...
+  onNetworkChange,
+}: SuperchainDropdownProps) => {
+  const [selectedNetwork, setSelectedNetwork] = useState<L2NetworkSelectItem>({
+    id: optimism.id,
+    chain: optimism,
+    config: superchain.op,
+  })
+
+  const allChains = useMemo(() => {
+    return Object.values(superchain).reduce((arr, config) => {
```
tremarkley, line 36: should this be filtered to only include l2s? I think with the way it is written now l1 eth would be included in this.

PR #290, opened 2024-04-23
add privy token to authorization header

nitaliano:
**Description**

Update the FE & BE to support an `Authorization` header flow. Our deploy previews aren't on the same domain as the backend so the cookie flow won't work on deploy previews, updating our auth flow to support pulling the token from localStorage and passing it up in the `Authorization` header will get us around this.


apps/dapp-console/app/helpers/apiClient.ts
```diff
@@ -6,15 +8,28 @@ import type { ApiV0 } from '@eth-optimism/dapp-console-api'
 
 import { ENV_VARS } from '@/app/constants/envVars'
 
+function parseAccessToken(accessToken: string) {
+  return JSON.parse(accessToken)
```
nitaliano, line 12: this removes the extra quotes and slashes that are getting added in localStorage

tremarkley (approved):

PR #303, opened 2024-04-27
feat(paymaster-proxy): create JsonRpcRouter

jakim929:
To improve maintainability, will refactor the paymaster-proxy logic to have cleaner separation between JsonRPC handling logic vs. business logic. the following helpers will be used. the refactor will be in a subsequent PR

### JsonRpcRouter

```typescript
const jsonRpcRouter = new JsonRpcRouter()

jsonRpcRouter.method(
  'subtract',
  z.tuple([z.number(), z.number()]),
  async ([a, b]) => a - b,
)

const result = await jsonRpcRouter.handle({
  jsonrpc: '2.0',
  method: 'subtract',
  params: [23, 42],
  id: 2,
}


// result
{ jsonrpc: '2.0', result: -19, id: 2 }

```

- as part of the refactor, introducing a new JsonRpcRouter class that helps set up rpc method handlers in an express-like way. 
- notice that individual method handler (`.method()` ) does not need to return a JSON rpc response, the logic to wrap it into a correct response is handled at the JsonRpcRouter level

### createExpressRouterFromJsonRpcRouter

creates an express router that handles json parsing + error handling.

### JsonRpcCastableError

- all errors encountered during the handlers should throw this type of error so that we can return the correct code + message.
- any other error thrown will just return a -32603 JSON rpc error
- note: JSON rpc calls will always try to return 200, even if the result is an error. Only exceptions that return 500 will be unexpected server errors

### DIFF
--- a/apps/paymaster-proxy/src/errors/JsonRpcCastableError.ts
+++ b/apps/paymaster-proxy/src/errors/JsonRpcCastableError.ts
@@ -1,0 +1,9 @@
+export abstract class JsonRpcCastableError extends Error {
+  override name = 'JsonRpcCastableError'
+  constructor(
+    readonly jsonRpcErrorCode: number,
+    readonly jsonRpcErrorMessage: string,
+  ) {
+    super(jsonRpcErrorMessage)
+  }
+}
--- a/apps/paymaster-proxy/src/jsonRpc/createExpressRouterFromJsonRpcRouter.ts
+++ b/apps/paymaster-proxy/src/jsonRpc/createExpressRouterFromJsonRpcRouter.ts
@@ -1,0 +1,41 @@
+import type { NextFunction, Request, Response } from 'express'
+import express from 'express'
+
+import { JsonRpcError } from '@/errors/JsonRpcError'
+import type { JsonRpcRouter } from '@/jsonRpc/JsonRpcRouter'
+
+export const createExpressRouterFromJsonRpcRouter = (
+  jsonRpcRouter: JsonRpcRouter,
+) => {
+  const router = express.Router()
+
+  router.use('/', (req, res, next) => {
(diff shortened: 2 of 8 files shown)


jakim929:
* **#303** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/303?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/paymaster-proxy/src/errors/JsonRpcCastableError.ts
```diff
@@ -0,0 +1,9 @@
+export abstract class JsonRpcCastableError extends Error {
```
jakim929, line 1: all errors handled during business logic will throw this a child of this error so that it can return the correct code / error message

apps/paymaster-proxy/src/jsonRpc/createExpressRouterFromJsonRpcRouter.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  router.use('/', (req, res, next) => {
+    // reject non-JSON requests
+    if (req.headers['content-type'] !== 'application/json') {
+      return res.status(400).send()
+    }
+    next()
+  })
+
+  router.use('/', express.json())
```
jakim929, line 20: Json RPC Parse error is handled here instead of at JsonRpcRouter since it is closely coupled to the bodyparsing logic in express. kept it here to prevent JsonRpcRouter from being tightly coupled with express.

another option is to just pass the raw buffer to the JsonRpcRouter and let it handle the parsing logic, but adds complexity so decided against in favor of just using express.json()

tremarkley (approved):

PR #306, opened 2024-04-30
Updates form to place errors under specific field or button

nitaliano:
**Description**

* Updates error handling for creating/verifying contracts, and claiming rebates to use custom codes returned from trpc api
* Updates finish verification dialog to use `react-form-hook` so we don't need to implement any extra logic for showing/hiding errors and having buttons disabled/enabled

### DIFF
--- a/apps/dapp-console/app/settings/components/ClaimRebateDialog.tsx
+++ b/apps/dapp-console/app/settings/components/ClaimRebateDialog.tsx
@@ -13,7 +13,7 @@ import { useCallback, useMemo, useState } from 'react'
 import { MAX_CLAIMABLE_AMOUNT } from '@/app/constants/rebate'
 import { shortenAddress } from '@eth-optimism/op-app'
 import Link from 'next/link'
-import { Contract } from '@/app/types/api'
+import { ApiError, Contract } from '@/app/types/api'
 import { Network } from '@/app/components/Network'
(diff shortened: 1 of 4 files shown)


apps/dapp-console/app/settings/components/ClaimRebateDialog.tsx
```diff
@@ -28,6 +28,13 @@ export type ClaimRebateDialogProps = {
   onRebateClaimed: (contract: Contract) => void
 }
 
+const errorMessages: Record<string, string> = {
```
tremarkley, line 31: nit: if you update this type to  `Partial<Record<DAPP_CONSOLE_ERROR_CODE, string>>`, then you will be able to make sure that the keys in this match the error codes

apps/dapp-console/app/settings/contracts/AddContractForm.tsx
```diff
@@ -169,7 +209,10 @@ export const AddContractForm = ({
         />
       </Form>
       <div className="mt-3 flex flex-row justify-between w-full">
-        <L2NetworkSelect onNetworkChange={handleNetworkChange} />
+        <L2NetworkSelect
+          includeTestnets
```
tremarkley, line 213: did you mean to leave `includeTestnets` in here?

tremarkley (approved):

PR #328, opened 2024-05-08
Updates feature flags to support labs emails

nitaliano:
**Description**

* Updates feature flags to support labs emails, you can now pass along a `allowDevs` option so even if the flag is disabled labs employees will still be able to see it and test it
* Renames `useFeature` -> `useFeatureFlag` to avoid export collision with growthbook
* Sends error to sentry if the feature flag value is not valid


jakim929 (approved):
nice! super useful

jvmi7 (approved):

PR #360, opened 2024-05-23
Github action to publish snapshots to NPM

nitaliano:
**Description**

Adds a new workflow called `release-snaphot` this will publish a snapshot of a **public** package to NPM. This is important to note that if your NPM package is marked as **private** publishing will be skipped even if a changeset exists for it. Now back to the conditions for getting snapshots published when we merge to our `main` branch.

* If a public package has never been published to NPM it will automatically create an entry
* If a changeset exists it will publish the package to NPM (Run `pnpm changeset` this will prompt you with the interactive cli)

This PR also adds/updates the following
* Adds `@changesets/changelog-github` as a root dev dep
* Marks all packages that we dont' want to publish as private. This is everything besides `op-app` and `sdk`

Example (tested this from a PR): https://www.npmjs.com/package/@eth-optimism/op-app

### DIFF
--- a/.changeset/flat-laws-cross.md
+++ b/.changeset/flat-laws-cross.md
@@ -1,0 +1,5 @@
+---
+"@eth-optimism/op-app": patch
+---
+
+Test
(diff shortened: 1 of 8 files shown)


.changeset/flat-laws-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+"@eth-optimism/op-app": patch
+---
+
+Test
```
nitaliano, line 5: I'm intentionally merging this in. Going to use this for publishing semver versions to NPM

jakim929 (approved):

PR #377, opened 2024-06-10
feat(contracts-ecosystem): add typescript dev dependency

tremarkley:


tremarkley:
* **#377** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/377?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jakim929 (changes requested):
thanks for fixing the build!

one rc is to create better separation between solidity / ts files. right now they both live in `src`. I think `src` should be purely for solidity code.

since the only reason to have typescript + ts build process in this package is to generate typescript bindings, maybe we can put it in a separate folder like `ts-abi` or `bindings` outside of the src, and export those instead

jakim929 (approved):

PR #422, opened 2024-07-26
update to node22

nitaliano:
**Description**

Starting to see some packages throughout the ecosystem move over to `node22`. Instead of having to polyfill any of the new functions that exist in it we should just bump our node version so we're already ready for when node18 support gets dropped in April


nickbalestra (approved):

nitaliano:
netlify doesn't have full support for 22 yet so going to leave this hanging around

PR #432, opened 2024-08-01
feat(dapp-console-api): use gelato transaction sender for faucet drip

jakim929:
- creates TransactionSender interface 
- create a gelato relay based tx sender
- use pseudorandom nonce rather than one based on account nonce
- TODO: update faucet route

Notes on faucet route API:
- this will cause a breaking change for the FaucetRoute API, since the async nature of TransactionSender means the endpoint will not be be able to synchronously return a tx hash
- instead of a breaking change, we should
    1. return an empty tx hash for the existing route `/offchainClaims` `/onChainClaims`, and etherscanUrl can just be to the etherscan website. Alternatively, we can use the createWalletTransactionSender so we can immediately return hash temporarily.
    2. create new routes `/triggerOffchainClaim` `/getOffchainClaimStatus`, `/triggerOnChainClaim` `/getOnChainClaimStatus`
    3. update front end to use new routes
    4. a week later, deprecate `/offchainClaims` `/onChainClaims` - the only consumer of the API is the web app so it shouldn't be too problematic

### DIFF
--- a/apps/dapp-console-api/src/Service.ts
+++ b/apps/dapp-console-api/src/Service.ts
@@ -23,6 +23,8 @@ import { optimism, optimismSepolia } from 'viem/chains'
 
 import type { ApiQueue } from '@/api-queue/apiQueueConfig'
 import { createApiQueue } from '@/api-queue/createApiQueue'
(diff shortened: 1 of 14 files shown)


Elgayar777 (approved):

Elgayar777 (commented):
- [ ]

PR #438, opened 2024-08-07
Setup shell for @eth-optimism/wagmi

nitaliano:
**Description**

* Fixes `@eth-optimism/viem` build
* Adds shell for `@eth-optimism/wagmi`
* Adds some base hooks we'll update and add tests for in a follow on PR
  * `useSendL2ToL2Message`
  * `useExecuteL2ToL2Message`
  * `useWaitForMessageIdentifier` 
* Removes `tsup` config files since we're only going to use `tsc`


packages/wagmi/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "react": "^18",
+    "resolve-tspaths": "^0.8.18",
+    "tsup": "^8.0.1",
+    "typescript": "^5.2.2",
+    "viem": "^2.17.9",
+    "vitest": "^1.6.0",
+    "wagmi": "^2.12"
+  },
+  "peerDependencies": {
+    "@eth-optimism/viem": "workspace:*",
```
nitaliano, line 36: once we start to version these this will be updated

packages/wagmi/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "lint:fix": "eslint \"**/*.{ts,tsx}\" --fix --quiet && pnpm prettier \"**/*.{ts,tsx}\" --write --loglevel=warn"
+  },
+  "devDependencies": {
+    "@eth-optimism/viem": "workspace:*",
+    "@tanstack/react-query": "^5",
+    "@types/react": "^18",
+    "@viem/anvil": "^0.0.7",
+    "react": "^18",
+    "resolve-tspaths": "^0.8.18",
+    "tsup": "^8.0.1",
```
tremarkley, line 29: can tsup be removed since we are not using it?

packages/viem/package.json
tremarkley, line 1: should tsup be removed as a devDependency now that we are not using it?

packages/wagmi/src/hooks/useWaitForMessageIdentifier.ts
```diff
@@ -0,0 +1,31 @@
...
+}: WaitForMessageIdentifierParams) => {
+  const config = useConfig()
+  const publicClient = usePublicClient({ config })
+  const { data } = useWaitForTransactionReceipt({ config, hash })
+
+  const id = useMemo<Promise<MessageIdentifier | undefined>>(async () => {
+    if (!data) {
+      return
+    }
+    return await extractMessageIdentifierFromLogs(
```
tremarkley, line 24: doesnt have to be done in this PR but, `extractMessageIdentifierFromLogs` will need to be updated now that L2toL2CrossDomainMessenger is not logging with the SentMessage event.

tremarkley: I also, wonder if we should extend `extractMessageIdentifierFromLogs` to also return the message payload, since that is also needed in order to execute a message.

tremarkley (approved):

packages/wagmi/src/hooks/useWaitForMessageIdentifier.ts
nitaliano: yea that was actually the intention to include the payload as well! i'll update it in the next PR i do

PR #453, opened 2024-09-04
adds initial changeset for viem & wagmi

nitaliano:
**Description**

Adds initial changeset to publish `0.0.1` version, snapshots are already being published on `main` merges


fainashalts (approved):
LGTM!

PR #467, opened 2024-09-09
feat(dapp-console-api): check that faucet user has github account linked

tremarkley:
Part of https://github.com/ethereum-optimism/ecosystem/issues/449

Server side changes to enforce that users of the faucet have a github account linked to their privy account. This also updates the github account information to be stored in the session instead of fetching it from privy on each request.

### DIFF
--- a/apps/dapp-console-api/src/middleware/isGithubAuthed.ts
+++ b/apps/dapp-console-api/src/middleware/isGithubAuthed.ts
@@ -1,0 +1,64 @@
+import { TRPCError } from '@trpc/server'
+
+import type { GrowthbookStore } from '@/growthbook'
+import { metrics } from '@/monitoring/metrics'
+import { Trpc } from '@/Trpc'
+
+/** Middleware used for checking if the request is from a privy user with a github account linked. */
+export const isGithubAuthed = (
+  trpc: Trpc,
+  growthBookStore: GrowthbookStore,
+) => {
+  return trpc.middleware(async ({ ctx, next }) => {
+    const { session } = ctx
+
+    if (!growthBookStore.get('enable_github_auth')) {
+      return next()
+    }
+
+    if (!session.user || !session.user.privyDid) {
+      throw new TRPCError({
+        code: 'UNAUTHORIZED',
+        message: 'User is not logged into privy',
+      })
+    }
+
+    if (!!session.user.githubSubject) {
+      return next()
+    }
+
+    const user = await trpc.privy
+      .getUser(session.user.privyDid)
+      .catch((err) => {
+        metrics.fetchPrivyUserErrorCount.inc()
+        trpc.logger?.error(
+          {
+            error: err,
+            entityId: session.user?.entityId,
+            privyDid: session.user?.privyDid,
+          },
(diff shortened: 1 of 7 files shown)


tremarkley:
* **#467** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/467?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/dapp-console-api/src/middleware/isGithubAuthed.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    if (!session.user || !session.user.privyDid) {
+      throw new TRPCError({
+        code: 'UNAUTHORIZED',
+        message: 'User is not logged into privy',
+      })
+    }
+
+    if (!!session.user.githubSubject) {
```
hamdiallam, line 26: what does this do?

tremarkley: if the github subject is already stored on the user's session it bypasses making a fetch to privy for the user's github info

apps/dapp-console-api/src/middleware/isGithubAuthed.ts
```diff
@@ -0,0 +1,64 @@
...
+            entityId: session.user?.entityId,
+            privyDid: session.user?.privyDid,
+          },
+          'error fetching privy user',
+        )
+        throw Trpc.handleStatus(500, 'error fetching privy user')
+      })
+
+    if (!user.github) {
+      throw new TRPCError({
```
nitaliano, line 46: not something for this PR but I think we should start moving our errors over to a single location so we can just call a function to get the error thrown with a description just so we don't have to rewrite the code

apps/dapp-console-api/src/middleware/isGithubAuthed.ts
```diff
@@ -0,0 +1,64 @@
...
+    session.save().catch((err) => {
+      metrics.failedToSaveUserIronSessionErrorCount.inc()
+      trpc.logger.error(
+        { err, entityId: session.user?.entityId },
+        'failed to save github subject to user iron session',
+      )
+      throw Trpc.handleStatus(500, 'unable to verify github auth')
+    })
+
+    return next()
```
nitaliano, line 62: is it guarantee'ed that `next()` won't be called before we end up hitting the `500` throw

nitaliano (approved):

apps/dapp-console-api/src/middleware/isGithubAuthed.ts
hamdiallam: why does this call `next()`? It seems like below it's updated

hamdiallam (approved):

apps/dapp-console-api/src/middleware/isGithubAuthed.ts
tremarkley: good catch, I missed adding an await to the `session.save` call

Issue #470, opened 2024-09-12
interop viem & wagmi v2


fainashalts:
The initial version of interop abstractions with viem and wagmi supports interop message passing. Now that the specs are clear for some of the asset transfer features (notably we are still waiting for L2-native SuperERC20), we can add asset transfer support.

Definition of Done: 

- [ ] viem actions for WETH interoperability
- [ ] wagmi hooks for WETH interoperability
- [ ] viem actions for L1-native SuperERC20
- [ ] wagmi hooks for L1-native SuperERC20
- [ ] viem actions for L2-native SuperERC20 (when ready)
- [ ] wagmi hooks for L2-native SuperERC20 (when ready)

fainashalts:
@nitaliano @tremarkley is there more to be done here? are we still planning to do something with `useSendSuperchainWETH`?

nitaliano:
We just opened a PR for it https://github.com/ethereum-optimism/ecosystem/pull/587 going to close it out

PR #479, opened 2024-09-20
Autogenerated ABI

nitaliano:
**Description**

Follow on to this https://github.com/ethereum-optimism/ecosystem/pull/476 which allows us to fetch the latest forge artifacts.

This PR includes the following
* Adds a `abigen.json` file that contains which contracts we want to autogenerate code for
* Script called `generate.ts` in our `viem` package. This will manage the generation of our `abis.ts` file
* Adds `eta` as a dev dependency, this is a simple template language that will allow us to auto generate code easier with templates.
* Includes the autogenerated abi code, the only real difference is now the file includes `SuperchainWETH`
* Updates `nx.json` to make `fetch:artifacts` a cacheable pre execution step before the `generate` commands happen

Tracker: https://github.com/ethereum-optimism/ecosystem/issues/470


packages/viem/src/abis.ts
tremarkley, line 1: just curious, is there a reason you prefer this to be in one file instead of a separate file for each abi?

packages/viem/package.json
```diff
@@ -14,13 +14,15 @@
     "typecheck": "tsc --noEmit --emitDeclarationOnly false",
     "typecheck:ci": "tsc --noEmit --emitDeclarationOnly false",
     "build": "tsc && resolve-tspaths",
+    "generate": "pnpm dlx tsx ./scripts/generate.ts",
```
tremarkley, line 17: nit: but would it be helpful to have a `fetchandgenerate` script as well that also runs the `fetch:artifacts` script since this is dependent on that?

tremarkley (approved):
thi

packages/viem/src/abis.ts
nitaliano: Right now all the abi in viem are in a single `abis.ts` file. I do think we're going to eventually update that and have everything in separate files once we start to try and autogenerate the entire protocol

packages/viem/package.json
nitaliano: This will also go in the direction of happening in CI as this evolves so we don't need to worry about running anything locally.

nitaliano: @tremarkley updated on this commit https://github.com/ethereum-optimism/ecosystem/pull/479/commits/32dd1210906a21dc59e63cbce4fe5f233f5b6205

This makes `fetch:artifacts` automatically happen when we attempt to run a generate command in a cacheable way. If we have `N` packages that all need to call `fetch:artifacts` it will only happen once and the others will read from the cache

PR #482, opened 2024-09-25
feat(faucet): implement rate limiting on faucet claims

tremarkley:
Part of https://github.com/ethereum-optimism/devx/issues/134

Implements rate limiting on faucet claim endpoints using redis cache.


tremarkley:
* **#482** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/482?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/dapp-console-api/src/middleware/faucetClaimRateLimiter.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    if (!ip) {
+      logger?.warn('faucet encountered a request with an undefined ip address')
+      return next()
+    }
+
+    const requestCount = await getFaucetClaimRate(redisCache, ip)
+    if (requestCount && requestCount >= BigInt(envVars.FAUCET_RATE_LIMIT)) {
+      logger?.warn('faucet request rate limited')
```
hamdiallam, line 35: Lets also log the ip address that is rate limited so that we have an idea of what's being limited

apps/dapp-console-api/src/constants/envVars.ts
```diff
@@ -91,6 +91,8 @@ const envVarSchema = z.object({
   API_QUEUE_REDIS_URL: z.string(),
   GROWTHBOOK_CLIENT_KEY: z.string(),
   GROWTHBOOK_ENCRYPTION_KEY: z.string().optional(),
+  FAUCET_RATE_LIMIT: z.number().min(1),
```
hamdiallam, line 94: IP_RATE_LIMIT, related places

apps/dapp-console-api/src/routes/faucet/FaucetRoute.ts
```diff
@@ -517,4 +548,14 @@ export class FaucetRoute extends Route {
       return 'GENERIC_ERROR'
     }
   }
+
+  private readonly updateFaucetClaimRate = async (ip?: string) => {
```
hamdiallam, line 552: I dont think we need to be checking the growthbook store many times here. It can be contractual that it is the callers responsibility to ensure this feature prior to calling this

apps/dapp-console-api/src/middleware/faucetClaimRateLimiter.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!growthBookStore.get('enable_faucet_rate_limit')) {
+      return next()
+    }
+
+    if (!ip) {
+      logger?.warn('faucet encountered a request with an undefined ip address')
+      return next()
+    }
+
+    const requestCount = await getFaucetClaimRate(redisCache, ip)
```
hamdiallam, line 33: Shouldn't we be incrementing the claim rate here? `INCR` updates and returns the latest value for the counter when updated. With this pattern we no longer need the `update`.

Simply something like
```javascript
const count = await incIPClaimRate(redisCache, ip)
if count > BigInt(envVars.FAUCET_RATE_LIMIT) {
   // rate limit
}
```

apps/dapp-console-api/src/routes/faucet/FaucetRoute.ts
```diff
@@ -226,6 +235,15 @@ export class FaucetRoute extends Route {
           authMode,
         }
       }
+      try {
+        await this.updateFaucetClaimRate(ctx.req.ip)
```
hamdiallam, line 239: why do this in the route when we can scope it to the middleware?

hamdiallam (commented):
instead of `faucetClaimRate`, can this be renamed to `faucetIPClaimRate` so it's a bit clearer what this rate limit is for?

apps/dapp-console-api/src/middleware/faucetClaimRateLimiter.ts
tremarkley: I intentionally am not incrementing here. we want to increment after a successful claim, so that is why I am doing the increment inside the route. I do not think they should be incremented strictly for hitting this endpoint, because there is a chance that the claim errors out due to an issue on our side and the user needs to retry it.

apps/dapp-console-api/src/routes/faucet/FaucetRoute.ts
tremarkley: see this comment: https://github.com/ethereum-optimism/ecosystem/pull/482/files#r1779308875

apps/dapp-console-api/src/middleware/faucetClaimRateLimiter.ts
hamdiallam: Event if it's a retry, the same RPC should be rate limited right? If there's transient error even on our side, continually spamming the retry will just make the problem worse. The client should back off a retry according to the overall rate limit on the endpoint

tremarkley: This should protect us from transient errors on our side. A claim is considered "successful" as soon as the tx hash is generated but before the transaction is confirmed on chain. This is where we are seeing issues, user's are continuing to retry to claim from the faucet while their claim tx is pending and has not gone through yet. By incrementing the rate limit here it allows us to know that the claim tx was created successfully and therefore we can more heavily rate limit this endpoint. Also something to keep in mind is we have a top level rate limiter that protects requests across all of our backend's endpoints.

apps/dapp-console-api/src/utils/faucetClaimRates.ts
```diff
@@ -0,0 +1,28 @@
+import { envVars } from '@/constants'
+
+import type { RedisCache } from './redis'
+
+const getFaucetRateLimitKey = (ip: string) => `faucet-rate-limit:${ip}`
+
+export const getFaucetClaimRate = async (redisCache: RedisCache, ip: string) =>
+  redisCache.getItem<bigint>(getFaucetRateLimitKey(ip))
+
+export const incrementFaucetClaimRate = async (
```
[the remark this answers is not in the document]
tremarkley: ok, this should be using the `incr` command now, good catch!

apps/dapp-console-api/src/middleware/faucetClaimRateLimiter.ts
hamdiallam: I see so the rate limit is on the txs created not the requests.

Curious why not just rate limit the top level requests for making the claim?
> This is where we are seeing issues, user's are continuing to retry to claim from the faucet while their claim tx is pending and has not gone through yet

This right here feels like the problem. We're only rate limiting when the claim goes through but will let callers attempt to create as many claim txs as it takes which will make the problem worse when overloaded. The top level rpc should instead be rate limited to begin this. 

In both scenarios, it's the top level call that should fail
1. Prolonged confirmation on-chain. Users increase rpc count for a successful claims -> rate limit
2. User tries to claim too much -> rate limite

tremarkley: > We're only rate limiting when the claim goes through but will let callers attempt to create as many claim txs as it takes which will make the problem worse when overloaded.

With the way this is currently implemented the rate limit will be incremented as soon as the tx is created. So in the case of prolonged confirmation on-chain we are still covered. 

I prefer to do the incrementation here instead of at the top level so that we can have a more strict rate limit, since we know that the tx has been created and is just waiting for confirmation on chain rather than a top level rate limit where an error might be due to something else and a retry could be valid.

hamdiallam: okay sounds good

hamdiallam (approved):

PR #496, opened 2024-10-09
fix(viem): separated L2ToL2CDM and CrossL2Inbox abstractions

hamdiallam:
With the latest changes to the CDM, the CrossL2Inbox & L2ToL2CDM abstractions can be completly seperated

1. Seperate action & utils between the two contracts
2. `Recipt:Messages` is `1:many` so the function interfaces should represent that when decoding the receipts
3. Relayed message decoding should cover both successful & failed messages
4. Add to the tests to make sure the cross chain calls are actually relayed (not failed)

Future Pull Request
- Viem actions to execute logs directly for users not using the L2CDM and directly consuming a message identifier & payload


hamdiallam:
* **#496** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/496?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @hamdiallam and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/viem/src/test/e2e/interop.spec.ts
```diff
@@ -1,84 +1,57 @@
...
-  chain: supersimL2B,
-  transport: http(RPC_URL_B),
-}).extend(walletActionsL2())
+import {
+  createInteropSentL2ToL2Messages,
+  decodeRelayedL2ToL2Messages,
+} from '@/utils/l2ToL2CrossDomainMessenger.js'
 
 describe('Generic Interop Flow', () => {
   it('should send and execute cross chain message', async () => {
```
nitaliano, line 19: i'm not sure why this test is failing in CI i'm starting to think theres some concurrency issue happening we can add `it.skip` to this so we can work on getting the PR merged in and come back to this and figure out whats happening

packages/viem/src/utils/interop.ts
```diff
@@ -0,0 +1,79 @@
...
+  const block = await client.getBlock({ blockHash: log.blockHash as Hash })
+  const id: MessageIdentifier = {
+    origin: log.address,
+    logIndex: BigInt(log.logIndex!),
+    blockNumber: block.number,
+    timestamp: block.timestamp,
+    chainId: BigInt(client.chain.id),
+  }
+
+  const payload = concat([...params.log.topics, params.log.data])
```
tremarkley, line 56: does the 0x have to be sliced off the front of each topic and log.data? something like:
```
function executingMessagePayloadBytes(log: {
  topics: string[]
  data: string
}): `0x${string}` {
  let msg = ''

  // Concatenate each topic (removing the "0x" prefix from each)
  for (const topic of log.topics) {
    msg += topic.startsWith('0x') ? topic.slice(2) : topic
  }

  // Concatenate log.data (removing the "0x" prefix if it exists)
  const logData = log.data.startsWith('0x') ? log.data.slice(2) : log.data

  // Return the concatenated result prefixed with "0x"
  return ('0x' + msg + logData) as `0x${string}`
}
```

packages/viem/src/test/e2e/interop.spec.ts
hamdiallam: confusing this is that this I think this is an error in the backend meaning it's hitting anvil

packages/viem/src/utils/interop.ts
hamdiallam: Yep it does, usage:   https://viem.sh/docs/utilities/concat.html#usage

nitaliano (approved):

packages/viem/src/utils/interop.ts
tremarkley: that's awesome, didnt know this existed

PR #615, opened 2025-01-14
viem/chains: chain exports for the superchain registry

### LINKED ISSUE #614: automatic viem chain definitions for all members in the registry
## Runtime Fetching
* Avoids the need for users to continually bump package versions as members are added to the superchain

* We don't want a github outage to result in an outage in superchain dapps. We can have 2/3 CDNs used as a fallback to github

## Compile time inclusion
 * To avoid complete runtime fetching, we could fallback to a maintained set of chains known at compile time. This set can be managed and consumers can just be encouraged to continually update their dependency but not strictly depend on it

hamdiallam:
**Description**

1. Introduce submodule dependency on the https://github.com/ethereum-optimism/superchain-registry
2. Add script to generate chains from the submodule. Exported under `viem/chains`

Generated chains follow the format `export const <name><network>`. i.e `opMainnet` or `baseSepolia`. If an app wants to add all chains, `export const mainnetChains` and `export const sepoliaChains` is a list of all the respective chains

**Tests**

N/A. But I'm thinking of adding tests to make sure the addresses match up with the ones specified in the `SystemConfig` for each network against the public rpc. However it's a safe property/assumption that the registry should already ensure that these are correct

**Metadata**

- Fixes #614 
    - Runtime fetching can be a followup. Not a needed issue for now


packages/viem/scripts/chaingen.ts
```diff
@@ -0,0 +1,144 @@
+import toml from '@iarna/toml'
+import { Eta } from 'eta'
+import fs from 'fs'
+import type { Address, PublicClient } from 'viem'
+import { createPublicClient, erc20Abi, http } from 'viem'
+import { mainnet, sepolia } from 'viem/chains'
+
+// Hardcoded. Can take a more elaborate approach if needed.
+const SUPERCHAIN_REGISTRY_PATH = '../../lib/superchain-registry'
```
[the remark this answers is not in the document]
hamdiallam: good shout! I'll update this

hamdiallam:
* **#619** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/619?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#615** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/viem/scripts/chaingen.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    const configPath = path.join(
+      SUPERCHAIN_REGISTRY_PATH,
+      'superchain',
+      'configs',
+      network,
+    )
+    const entries = fs
+      .readdirSync(configPath)
+      .filter((entry) => !entry.includes('superchain'))
```
tremarkley, line 99: just curious, why is this needed? might be good to add a comment here

packages/viem/scripts/chaingen.ts
```diff
@@ -0,0 +1,159 @@
...
+    const chainDefs = await Promise.all(
+      entries.map(async (entry) => {
+        const chainConfig = toml.parse(
+          fs.readFileSync(`${configPath}/${entry}`, 'utf8'),
+        )
```
tremarkley, line 104: nit: you could add a type here for the result of toml.parse so you can avoid having to cast the types each time you access something inside `chainConfig`

packages/viem/scripts/chaingen.ts
```diff
@@ -0,0 +1,159 @@
...
+        if (addresses.DisputeGameFactoryProxy) {
+          l1Addresses['disputeGameFactory'] = addresses.DisputeGameFactoryProxy
+        }
+        if (addresses.L2OutputOracleProxy) {
+          l1Addresses['l2OutputOracle'] = addresses.L2OutputOracleProxy
+        }
+
+        const normalizedName = entry
+          .replace('.toml', '')
+          .replace('-testnet', '')
```
tremarkley, line 127: if you dont want to rely on the file name for this, there is also the `identifier` field on chainList.json: https://github.com/ethereum-optimism/superchain-registry/blob/e2d3490729b20a649281899c2c286e6e12db57f3/chainList.json#L59. That could potentially be more stable.

tremarkley (approved):
nice job!

packages/viem/scripts/chaingen.ts
hamdiallam: every network dir within the `config/<network>` has a `superchain.toml` file. This file has the L1 superchain config for that nework

def worth a comment

packages/viem/scripts/chaingen.ts
hamdiallam: ahh yea this is a good callout. Need to be able to map this identifier to the filename in order to parse the address list

packages/viem/scripts/chaingen.ts
hamdiallam: It looks like they keep the same filepath structure within `config`. I'm going to leave it as it is right now but we can adjust later

PR #674, opened 2025-03-04
chore: update Node.js version to 18.18.0

raffaele-oplabs:
**Description**

This pull request includes a small change to the `mise.toml` file. The change downgrades the Node.js version from `20.11.0` to `18.18.0`.

Fixing error in CI:
Screenshot 2025-03-04 at 22 28 33

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,5 +1,5 @@
 [tools]
-node = "18.18.1"
+node = "18.18.0"
 
 "ubi:ethereum-optimism/supersim"="0.1.0-alpha.38"
 
--- a/mise.toml
+++ b/mise.toml
@@ -1,5 +1,5 @@
 [tools]
-node = "20.11.0"
+node = "18.18.0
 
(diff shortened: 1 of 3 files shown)


janjakubnanista (approved):

mise.toml
```diff
@@ -1,5 +1,5 @@
 [tools]
-node = "20.11.0"
+node = "18.18.0"
```
tremarkley, line 2: why is this change needed? would be helpful to have the context documented

raffaele-oplabs: The version used previously was not the right one.

Screenshot 2025-03-04 at 22 28 33

tremarkley, line 2: nit: can you make this 18.18.1 so that it aligns with https://github.com/ethereum-optimism/ecosystem/blob/main/.nvmrc?

tremarkley (approved):

PR #707, opened 2025-03-24
chore: nx single version policy & package cleanup

hamdiallam:
We're moving to single version policy for this monorepo
- All dev dependencies are shared
- Regular deps must be listed in root package.json and match locally as well

There's a `@nx/dependency-checks` ESLint rule to help validate the above. I'm going to get it working in a followup after eslint is migrated

This PR also removes the playground within ui-components package which is unused


hamdiallam:
* **#707** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/707?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/707?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/wagmi/package.json
```diff
@@ -20,23 +20,17 @@
     "typecheck": "tsc --noEmit --emitDeclarationOnly false",
     "typecheck:ci": "tsc --noEmit --emitDeclarationOnly false"
   },
-  "devDependencies": {
-    "@eth-optimism/viem": "workspace:^0.4.0",
+  "dependencies": {
```
tremarkley, line 23: why were these converted from `devDependencies` to `dependencies`? eth-optimism/wagmi should generally use `peerDependencies` for shared runtime dependencies like `react` and `wagmi`, while using `dependencies` only for packages that aren't expected to be provided by the consumer.

hamdiallam: it was made a dependency since wagmi directly uses viem in internally.

The consumer of wagmi does not have to install viem which i what I believe peerDependencies enforces when I was looking into it.

tremarkley: I think we should convert these back to being `devDependencies`. I believe if react, wagmi, and @tanstack/react-query are listed as dependencies, then @eth-optimism/wagmi is going to be built pinned to a specific version of react, wagmi, and @tanstack/react-query, which will potentially lead to conflicting versions of these frameworks if someone consumes @eth-optimism/wagmi, in their app and they use a different version of one of these frameworks.

packages/viem/package.json
```diff
@@ -21,25 +21,13 @@
...
-    "@iarna/toml": "^2.2.5",
-    "eta": "^3.1.1",
-    "resolve-tspaths": "^0.8.18",
-    "typedoc": "^0.26.7",
-    "typedoc-plugin-markdown": "^4.2.8",
-    "typescript": "^5.2.2",
-    "viem": "^2.17.9",
-    "vitest": "^1.6.0"
+  "dependencies": {
+    "viem": "^2.17.9"
```
tremarkley, line 30: similar comment here, I dont think we want to enforce a specific version of viem as a dependency

packages/wagmi/package.json
hamdiallam: devDependencies are dependencies not required at runtime (eslint, typescript, etc).

Since wagmi imports react, it has to be included as a dependency. However since it's also a peer dependency, it's still marked as compatible with higher versions

hamdiallam: https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

>"devDependencies": Packages that are only needed for local development and testing.

packages/viem/package.json
hamdiallam: we have to since this package uses viem as a dependency

hamdiallam: actually for this we should be able to leverage peerDependencies. This is different than the devDependencies. Let me revisit it

packages/wagmi/package.json
tremarkley: hmm, wagmi and react-query both use the pattern of not specifying dependencies. here are the dependencies for [@tanstack/react-query](https://github.com/TanStack/query/blob/main/packages/react-query/package.json), [wagmi core](https://github.com/wevm/wagmi/blob/main/packages/core/package.json), [wagmi-react](https://github.com/wevm/wagmi/blob/main/packages/react/package.json). viem and react are listed as peerDependencies and devDependencies.

The dependies + peerDependencies approach might still work because viem, react-query, react, and wagmi are also marked as a peerDependencies, but I would think we would want to have devDependencies + peerDependencies. devDependencies being for local development / tests and then peerDependencies for dependency resolution when the library is consumed. Is there a reason the dependencies + peerDependencies approach is preferred here?

packages/viem/package.json
hamdiallam: updated! good catch

packages/wagmi/package.json
hamdiallam: moved them to `peerDependencies` instead

packages/wagmi/package.json
```diff
@@ -12,31 +12,16 @@
...
     "@tanstack/react-query": "^5",
-    "@types/react": "^18",
-    "@viem/anvil": "^0.0.7",
     "react": "^18",
-    "resolve-tspaths": "^0.8.18",
-    "typescript": "^5.2.2",
-    "viem": "^2.17.9",
-    "vitest": "^1.6.0",
-    "wagmi": "^2.12"
-  },
```
tremarkley, line 34: same comment here, we should add these alongside the `peerDependencies`

packages/viem/package.json
```diff
@@ -21,25 +21,10 @@
...
-    "typedoc": "^0.26.7",
-    "typedoc-plugin-markdown": "^4.2.8",
-    "typescript": "^5.2.2",
-    "viem": "^2.17.9",
-    "vitest": "^1.6.0"
+    "lint": "eslint \"**/*.{ts,tsx}\" && pnpm prettier --check \"**/*.{ts,tsx}\" --ignore-path \"../../.prettierignore\"",
+    "lint:fix": "eslint \"**/*.{ts,tsx}\" --fix && pnpm prettier \"**/*.{ts,tsx}\" --write --loglevel=warn --ignore-path \"../../.prettierignore\"",
+    "test": "vitest --run",
+    "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
```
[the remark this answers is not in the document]
hamdiallam: ohh interesting, okay I'll add it as a dev in the rootmost package.json

package.json
```diff
@@ -20,20 +20,39 @@
...
   },
-  "dependencies": {
-    "@changesets/cli": "^2.27.1"
+  "peerDependencies": {
+    "viem": "^2.17.9",
+    "wagmi": "^2.12",
+
+    "react": "^18",
+    "@tanstack/react-query": "^5"
   }
```
tremarkley, line 57: should these all be devDependencies? I dont think we want any peerDependencies at the root, since the root of this repo is not a package that will be installed by any consumers.

tremarkley (approved):
🐐

PR #752, opened 2025-04-24
chore: start ponder-interop from env

hamdiallam:
For people running this app against a seperate network, we dont want them to edit the code for config.

So the default config is bootstrapped from env.
- dev:(alphanet|devnet|supersim) command to autostart known networks.
- Consolidate ponder config generation between the different ways to configure this.

Aside:
- add the url to the chains api so other apps can just bootstrap from this (autorelayer)
- fix api responses

### DIFF
--- a/apps/ponder-interop/src/createPonderConfig.ts
+++ b/apps/ponder-interop/src/createPonderConfig.ts
@@ -1,0 +1,44 @@
+import {
+  contracts as addrs,
+  l2ToL2CrossDomainMessengerAbi,
+} from '@eth-optimism/viem'
+import { createConfig } from 'ponder'
+import type { Transport } from 'viem'
+
+export type Endpoint = { chainId: number; transport: Transport }
+
+/**
+ * A map of name to {@link Endpoint}
+ */
+export type Endpoints = Record<string, Endpoint>
+
+/**
+ * Create a Ponder config for interop
+ * @param endpoints - Interoperable endpoints to index -- {@link Endpoints}
+ * @returns Ponder configuration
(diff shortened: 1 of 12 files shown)


hamdiallam:
* **#752** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/752?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ethereum-optimism/ecosystem/752?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/ponder-interop/src/createPonderConfig.ts
```diff
@@ -0,0 +1,44 @@
...
+export function createPonderConfig(endpoints: Endpoints) {
+  if (Object.keys(endpoints).length === 0) {
+    throw new Error('no endpoints provided')
+  }
+
+  // relevant interop contracts
+  const contracts = {
+    L2ToL2CDM: {
+      abi: l2ToL2CrossDomainMessengerAbi,
+      startBlock: 1,
```
tremarkley, line 29: side question - will it be possible to modify this on a per chain basis. it should ideally be equally to the block where interop first started on the chain, right? everything will still work if it is set to 1, but might be nice for performance if it can be set to a later value.

hamdiallam: ahh yea that's right. When we get to launch, we can make this configurable as a parameter

PR #813, opened 2025-05-14
feat: superchain-playground app

hamdiallam:
Repository: https://github.com/jakim929/superchain-tools


package.json
tremarkley, line 1: did you add these at the root level so that future apps that have a shadcn ui can share these dependencies?

tremarkley (approved):

package.json
hamdiallam: yep i just added them right when you reviewed I think.

Some typecheck errors I have to fix though

PR #818, opened 2025-05-15
feat(viem): network abstraction

hamdiallam:
### TL;DR

- Added Network type and utilities to organize chains by network
- added dotenv dependency to viem packages.

When interop launches, we'll likely have `sepolia-interop` and `mainnet-interop` networks which is a subset of these respective networks.


packages/viem/src/chains/chains.ts
```diff
@@ -0,0 +1,13 @@
+import type { Chain } from 'viem'
+
+import { networks } from '@/chains/networks.js'
+
+/**
+ * Map of all viem chains by ID
+ */
+export const chainById = Object.values(networks)
```
tremarkley, line 8: nice

tremarkley (approved):
lgtm!

PR #833, opened 2025-06-10
feat(ponder-interop): index `GasTank` events

tremarkley:
Closes https://github.com/ethereum-optimism/ecosystem/issues/838

## Changes
- Added new tables to track gas provider operations and message claims:
  - `gasProviders`: Tracks gas provider balances
  - `gasProviderPendingWithdrawals`: Tracks pending withdrawals
  - `flaggedMessages`: Tracks flagged messages
  - `claimedMessages`: Tracks claimed messages

These changes add support for tracking gas provider operations and message claims in the interop system. The first use case for this will be in the relayer where it will be used for querying gas provider balances prior to relaying messages.

### DIFF
--- a/apps/ponder-interop/ponder.schema.ts
+++ b/apps/ponder-interop/ponder.schema.ts
@@ -1,7 +1,7 @@
 // Hack to resolve https://github.com/ponder-sh/ponder/issues/1722
 import 'drizzle-orm/pg-core'
 
-import { index, onchainTable } from 'ponder'
+import { index, onchainTable, primaryKey } from 'ponder'
 
 export const sentMessages = onchainTable(
   'l2_to_l2_cdm_sent_messages',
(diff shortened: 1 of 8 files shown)


apps/ponder-interop/src/createPonderConfig.ts
```diff
@@ -36,6 +39,16 @@ export function createPonderConfig(endpoints: Endpoints) {
         ]),
       ),
     },
+    GasTank: {
```
hamdiallam, line 42: Lets omit this field if the env variable is empty?

apps/ponder-interop/ponder.schema.ts
```diff
@@ -52,3 +52,73 @@ export const relayedMessages = onchainTable(
     transactionHash: t.hex().notNull(),
   }),
 )
+
+export const gasTankGasProviders = onchainTable(
```
hamdiallam, line 56: since we support withdrawals, lets start with just eth_call the contract instead of indexing every deposit?

Would reduce the complexity

hamdiallam (approved):
Looks good! 2 comments

apps/ponder-interop/src/createPonderConfig.ts
tremarkley: yeah, that was the path I initially went down, but ponder types do not play well with the conditional GasTank. Let me check with the ponder team on whether this is supported

tremarkley: havent heard back from ponder team, but this works as-is so going to merge for now and follow up if their recommended approach is different

Issue #731, opened 2025-04-07
Add Standard Actions to @eth-optimism/viem docs


tremarkley:
Right now the docs for the standard actions inside the @eth-optimism/viem package are not being auto-generated. With this ticket implemented, running `pnpm run docs` inside of the viem package, should autogenerate docs for the non-interop actions in the @eth-optimism/viem package

PR #859, opened 2025-06-16
feat: Include non-interop actions in docs generation

its-everdred:
## Problem
[issue](https://github.com/ethereum-optimism/ecosystem/issues/731)
Running `pnpm run docs` inside of the viem package does not auto-generate docs for the non-interop actions in the `@eth-optimism/viem` package

## Solution
Include these actions in `typedoc.json` and generate new docs.


packages/viem/typedoc.json
```diff
@@ -1,16 +1,17 @@
...
-        "title.memberPage": "{name}",
-    }
-}
\ No newline at end of file
+  "$schema": "https://typedoc.org/schema.json",
+  "out": "./docs",
+  "plugin": ["typedoc-plugin-markdown"],
+  "entryPoints": [
+    "./src/index.ts",
+    "./src/actions/index.ts",
```
its-everdred, line 7: This line is the only real addition. Everything else in this file is prettier, everything else in this PR is the newly generated docs

PR #869, opened 2025-06-24
feat(ponder-interop): api endpoint for fetching pending GasTank claims

tremarkley:
Part of https://github.com/ethereum-optimism/ecosystem/issues/840

## Changes 
- Updates the `gas_tank_relayed_message_receipts` to store the message identifier and log payload of the `RelayedMessageGasReceipt` event
- Adds the `/messages/pending/claims` endpoint to the ponder api, which returns a list of relayed messages that should be claimed from the `GasTank`

### DIFF
--- a/apps/ponder-interop/src/api/index.ts
+++ b/apps/ponder-interop/src/api/index.ts
@@ -12,7 +12,8 @@ import {
 } from 'ponder'
 import { db, publicClients } from 'ponder:api'
 import schema from 'ponder:schema'
-import { isAddress } from 'viem'
+import type { Address } from 'viem'
+import { getAddress, isAddress } from 'viem'
 
 type GasProvider = {
   gasTankChainId: number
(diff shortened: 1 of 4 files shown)


apps/ponder-interop/src/api/index.ts
```diff
@@ -198,3 +300,5 @@ function getPendingMessagesQuery() {
     )
     .orderBy(desc(schema.sentMessages.timestamp))
 }
+
+export default app
```
its-everdred, line 304: 👍 `ponder-interop` is small enough that it's still convenient to have all these endpoints in one file. we may consider breaking them into their own files as we add more though

its-everdred (approved):
👍 one nit, but looks good

apps/ponder-interop/src/api/index.ts
tremarkley: yeah, definitely something we should consider as this grows!

PR #870, opened 2025-06-24
feat(autorelayer-interop): claim from GasTank

tremarkley:


apps/autorelayer-interop/src/relayer.ts
```diff
@@ -45,16 +45,16 @@ export type PendingMessage = z.infer<typeof PendingMessageSchema>
 
 export type PendingMessages = z.infer<typeof PendingMessagesSchema>
 
+const gasTankProviderSchema = z.object({
+  gasTankChainId: z.number(),
+  gasProviderBalance: z.number(),
+  gasProviderAddress: z
+    .string()
+    .refine(isAddress, 'invalid gas provider address'),
+})
```
its-everdred, line 54: 🤯 TIL about `zod`. saves so much manual validation

apps/autorelayer-interop/src/relayer.ts
```diff
@@ -205,6 +155,258 @@ export class Relayer {
...
+
+    // If no account is configured (sponsored), specify
+    // a random account owned by the sponsored endpoint.
+    let account = walletClient.account
+    if (!account) {
+      const addresses = await walletClient.getAddresses()
+      if (addresses.length === 0) {
+        msgLog.warn('no accounts found, skipping...')
+        return null
+      }
```
its-everdred, line 321: ❓ If I'm running a relayer, what's the use-case for using many addresses? Redundancy?

its-everdred (approved):
👍

hamdiallam (approved):

apps/autorelayer-interop/src/relayer.ts
tremarkley: I added a better explanation of this to the comment, but this is only needed for typescript purposes. In the sponsored case viem will send a `eth_sendTransaction` RPC to the sponsored sender service, and the sponsored sender service will handle signing and submitting the transaction

PR #887, opened 2025-07-07
feat(autorelayer-interop): incorporate pending claims into GasProvider balance calculations

tremarkley:
Closes https://github.com/ethereum-optimism/ecosystem-private/issues/398

## Changes
- When calculating the balance for a given `GasProvider` deducts any unclaimed relays from the `GasProvider`s balance

### DIFF
--- a/apps/autorelayer-interop/src/relayer.ts
+++ b/apps/autorelayer-interop/src/relayer.ts
@@ -103,6 +103,21 @@ export type PendingClaim = z.infer<typeof PendingClaimSchema>
 
 export type PendingClaims = z.infer<typeof PendingClaimsSchema>
 
+const PendingRelayCostForGasProviderSchema = z
+  .object({
+    gasProviderAddress: z
+      .string()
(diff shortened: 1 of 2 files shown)


apps/autorelayer-interop/src/relayer.ts
```diff
@@ -294,28 +319,62 @@ export class Relayer {
...
-            ? provider
-            : best,
-        message.gasTankProviders[0],
-      )
+      for (const provider of message.gasTankProviders) {
+        const balance = await getCachedBalance(provider)
+        if (balance > bestGasProvider.gasProviderBalance) {
+          bestGasProvider = provider
+        }
+      }
```
its-everdred, line 368: 🤔 Is it odd that the "bestGasProvider" may not have enough balance to relay the message? We could that the provider has enough balance to send the message, but it would make the response object a little more complicated. Maybe it's not worth worrying about?

its-everdred (approved):

PR #908, opened 2025-07-16
Add wallet integration to verbs-sdk

its-everdred:
## Problem

The new verbs-sdk needs a wallet integration

## Solution

- [x] Establish verbs-sdk config, types
- [x] Establish Wallet class
- [x] Integrate with Privy as first wallet adapter
- [x] Extend verbs-service with new sdk and wallet functions
- [x] Add typedocs and comments

### DIFF
--- a/packages/verbs-sdk/src/adapters/privy.ts
+++ b/packages/verbs-sdk/src/adapters/privy.ts
@@ -10,7 +10,7 @@ import { Wallet } from '../wallet.js'
  */
 export class PrivyWalletProvider implements WalletProvider {
   private privy: PrivyClient
-  private chainId: number = 10 // TODO: make configurable
+  private chainId: number = 130 // TODO: make configurable
 
   /**
    * Create a new Privy wallet provider
@@ -36,7 +36,6 @@ export class PrivyWalletProvider implements WalletProvider {
 
       return new Wallet(wallet.id, wallet.address as Address, this.chainId)
     } catch (error) {
-      console.error('Error creating wallet:', error)
       throw new Error(`Failed to create wallet for user ${userId}`)
     }
   }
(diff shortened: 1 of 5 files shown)


packages/verbs-sdk/src/verbs.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  private walletProvider: WalletProvider
+
+  constructor(config: VerbsConfig) {
+    this.walletProvider = this.createWalletProvider(config)
+
+    // Delegate wallet methods to wallet provider
+    this.createWallet = this.walletProvider.createWallet.bind(this.walletProvider)
+    this.getWallet = this.walletProvider.getWallet.bind(this.walletProvider)
+  }
```
its-everdred, line 22: 🎨 We may decide this is over-engineered, but my intention is to establish this pattern: The Verbs SDK delegates functionality to various provider abstractions, which may have various implementations.

packages/verbs-sdk/src/adapters/privy.ts
```diff
@@ -0,0 +1,59 @@
...
+import type { WalletProvider } from '../types/wallet.js'
+import { Wallet } from '../wallet.js'
+
+/**
+ * Privy wallet provider implementation
+ * @description Wallet provider implementation using Privy service
+ */
+export class PrivyWalletProvider implements WalletProvider {
+  private privy: PrivyClient
+  private chainId: number = 10 // TODO: make configurable
```
fainashalts, line 13: agree eventually we want this to be configurable but in the meantime we may want to set this to 130 as its Unichain's chainId and the pool we're going to demo is there.

packages/verbs-sdk/src/verbs.ts
fainashalts: I think this makes sense. we can revisit when we design the MVP though!

fainashalts (approved):
LGTM!

Issue #906, opened 2025-07-15
Add wallet to verbs


its-everdred:
## Problem

Verbs requires a wallet package

## Solution

Integrate with Privy for wallet creation

PR #914, opened 2025-07-17
Add getAllWallets to verbs

its-everdred:
## Problem
[ticket](https://github.com/ethereum-optimism/ecosystem/issues/906)
We don't have a way to query wallets in verbs.

## Solution

- [x] Add `getAllWallets` to sdk
- [x] Add `GET /wallets` to service
- [x] Add `wallets list` to ui


apps/verbs-ui/src/components/Terminal.tsx
```diff
@@ -75,7 +79,7 @@ const Terminal = () => {
       {
         id: 'welcome-8',
         type: 'output',
-        content: '    Web2 -> Web3 library for the OP Stack',
+        content: '   Money Verbs library for the OP Stack',
```
fainashalts, line 82: Not super tied to this in either direction but I don't know that this needs the "Money" -- just "Verbs library" seems fine to me.

apps/verbs-ui/src/components/Terminal.tsx
```diff
@@ -123,6 +129,23 @@ const Terminal = () => {
     return data
   }
 
+  const getAllWallets = async (): Promise<GetAllWalletsResponse> => {
+    const response = await fetch('http://localhost:3000/wallets', {
```
fainashalts, line 133:
```suggestion
    @TODO abstract the url here
    const response = await fetch('http://localhost:3000/wallets', {
```

fainashalts (approved):
LGTM, left some small suggestions. I see automation found a possible security finding so be sure to take a look at that before merging in case its anything!

apps/verbs-ui/src/components/Terminal.tsx
its-everdred:
```suggestion
        content: '   Verbs library for the OP Stack',
```
Makes sense

